PR #85, opened 2021-11-02
Fe 100 create advice carousel

BenHakimIlyass:
- Removed all carousel external packages.
- Created carousel using `ScrollView`


yarn.lock
```diff
@@ -1601,6 +1601,46 @@
     dequal "^1.0.0"
     isoworker "^0.1.1"
 
+<<<<<<< HEAD
```
natsukiai, line 1604: Is that normal ? `yarn install` doesn't throw any error, so...

natsukiai (approved):
Tested on web and android. It works. :-)
Be sure to test on iOS and send pictures of you work on iOS.

yarn.lock
BenHakimIlyass: I believe this is because I added the `react-native-swiper-flatlist` and them I removed it.

PR #87, opened 2021-11-03
:rocket: Replace .gif advice assets by .webp to avoid performance dro…

BenHakimIlyass:
## Webp

As I was testing the my latest changes on Android, I noticed a big drop of performance, and that was because of the heavy gif images.

I replaced the `.gif` images by `.webp`, it works way better on IOS, web and especially on Android.

image

💡  We still can compress these images and get a smaller size.


natsukiai (approved):
Great job Ilyass!

PR #88, opened 2021-11-03
Fixing media queries, not for notch devices

BenHakimIlyass:
### CameraView

The CameraView now supports all screen sizes, by transforming the screen `width` and `height` into a ratio.
Works on all platforms ✅ .

### Notch issues

As we know when getting the screen dimensions, we're getting the whole height and width including the **notch** size, which we have to remove from the ratio calculation.


natsukiai (commented):
Files are good to me, but I have still some issues on iOS for example.
Can you maybe draw a small schema on how the queries are working to help understand the all compatibility ?

BenHakimIlyass:
Will do.

BenHakimIlyass:
## New updates

### The problem

When running the camera view on normal phones, technically we generate the Camera ratio based on the screen width which is not a problem:

image

But when running the camera view on a phone with a bigger status bar (notch), the size of this notch will not be calculated in the camera ratio.

image

## Solution

Instead of generating this ratio based on the screen width, we can use the `safeViewWidth` as:

image

natsukiai (approved):
Great job 👏

PR #92, opened 2021-11-12
Refactor Camera hooks and views

natsukiai:


natsukiai (commented):
Seems all good to me after rechecking.

BenHakimIlyass:
Looks good to me as well, @natsukiai 🥂.

website/docs/js/api/react-native-views.md
```diff
@@ -31,7 +31,7 @@ import { CameraView } from '@monkvision/react-native-views';
 ```
 
 ### Response payload
-`{ pictures, camera, sights }`
+`{ pictures, camera, sights, nbOfSights }`
 ``` json
 // const handleSuccess = (payload) => console.log(payload);
 // <CameraView onSuccess={handleSuccess} />
```
franklinkemta, line 37: \O-O\

franklinkemta (approved):

BenHakimIlyass (approved):

packages/react-native-views/src/components/CameraView/CameraPopUps/index.jsx
```diff
@@ -0,0 +1,66 @@
...
+  onDismissAdvices,
+  onDismissSnack,
+  snackIsVisible,
+}) {
+  const { colors } = useTheme();
+
+  return (
+    <>
+      <Modal
+        contentContainerStyle={styles.advices}
```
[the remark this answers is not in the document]
BenHakimIlyass: True

website/docs/js/api/react-native-views.md
natsukiai: This is a markdown file. ^^

packages/react-native-views/src/components/CameraView/hooks/usePictures.js
```diff
@@ -0,0 +1,35 @@
...
+      const picture = await camera.takePictureAsync(options);
+      const payload = { sight: activeSight, source: picture };
+
+      setPictures((prevState) => ({ ...prevState, [activeSight.id]: payload }));
+      onTakePicture(payload);
+
+      if (!nextSightProps.disabled) {
+        nextSightProps.onPress();
+      }
+    }
```
rb46546741, line 24: I usually like to change this kind of patterns into 
```js
if (!camera) return;

...
```
It reduces indentation and clearly shows that not having a camera instantly exits.

packages/react-native-views/src/components/CameraView/index.jsx
```diff
@@ -34,187 +59,81 @@ export default function CameraView({
...
+              {ui.container.measures.width && (
+                <Components.Camera
+                  onCameraReady={handleCameraReady}
+                  ratio={ratio}
                 />
               )}
+              <CameraOverlay
+                activeSightId={activeSight.id}
+                camera={camera}
+                fakeActivity={Boolean(fakeActivity)}
```
rb46546741, line 114: Why not make sure it's a boolean in the `useFakeActivity` hook directly?

packages/react-native-views/src/components/TutorialView/Illustration.jsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+
+export default () => (
+  <svg xmlns="http://www.w3.org/2000/svg" data-name="Layer 1" height="220px" viewBox="0 0 858.756 581.934">
```
rb46546741, line 4: We loose the fact that this is a svg file (with all the advantages, such as the ability to preview and edit it as such) and it becomes hard to tell if the native and non native file are the same or not.
Wouldn't it be better to have an actual svg file and manipulate it?

src/screens/InspectionCreate/index.jsx
```diff
@@ -3,23 +3,21 @@ import React, { useCallback } from 'react';
 import { CameraView } from '@monkvision/react-native-views';
 import { useNavigation } from '@react-navigation/native';
 
-import { TUTORIAL } from 'screens/names';
+import { GET_START, INSPECTION_REVIEW } from 'screens/names';
```
rb46546741, line 6: GET_START isn't the greatest name in the world ;) Is it a "getting started" screen?

packages/react-native-views/src/components/CameraView/CameraPopUps/index.jsx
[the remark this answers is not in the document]
natsukiai: Can one of you both make a fix for that ? I have delivery tasks on the bench.

packages/react-native-views/src/components/CameraView/index.jsx
```diff
@@ -1,18 +1,43 @@
...
+    justifyContent: 'space-between',
+    ...Platform.select({
+      native: { flex: 1 },
+      default: { display: 'flex', flex: 1 },
+    }),
+  },
+});
 
 const SIDEBAR_WIDTH = 250;
 const makeRatio = (width, height) => `${(width - SIDEBAR_WIDTH) / 240}:${height / 240}`;
```
rb46546741, line 43: Why not put the SIDEBAR_WIDTH in the style (and make `make_ratio` have it as a parameter)

The hardcoded 240 seems fishy :suspect:

packages/react-native-views/src/components/CameraView/hooks/useFakeActivity.js
```diff
@@ -0,0 +1,23 @@
...
+import noop from 'lodash.noop';
+
+function useFakeActivity() {
+  const [fakeActivity, setFakeActivity] = useState(null);
+
+  const handleFakeActivity = useCallback((onEnd = noop) => {
+    const fakeActivityId = setTimeout(() => {
+      setFakeActivity(null);
+      onEnd();
+    }, 500);
```
rb46546741, line 11: Why 0.5 seconds ? This should probably have a variable name and be just after the imports.

packages/react-native-views/src/components/CameraView/hooks/useSuccess.js
```diff
@@ -0,0 +1,15 @@
+import { useEffect } from 'react';
+
+function useSuccess(onSuccess, payload, handleFakeActivity) {
+  const { pictures, camera, sights } = payload;
+  const nbOfSights = sights.length;
+
+  useEffect(() => {
+    const picturesTaken = Object.values(pictures).filter((p) => Boolean(p.source)).length;
```
rb46546741, line 8: Doesn't `filter` already coerce the result of the function to a boolean?

If we use `nbOfSights` shouldn't we use `nbOfPicturesTaken` here? (I personally prefer `sightsCount` and `takenPicturesCount`)

packages/react-native-views/src/components/CameraView/hooks/useUI.js
```diff
@@ -0,0 +1,59 @@
+import { useCallback, useState } from 'react';
+
+function useUI(camera, pictures, onCloseCamera, onShowAdvice) {
```
rb46546741, line 3: I don't understand what this hooks does from the name :smiling_face_with_tear:

packages/react-native-views/src/components/CameraView/index.jsx
natsukiai: Because fakeActivity is a timeoutId and if we want to break the activity, we can juste use `clearTimeout(fakeActivity)`.

packages/react-native-views/src/components/TutorialView/Illustration.jsx
natsukiai: It's complicated because svg files are not loaded the same way in every projects. Between the default Metro config or the default Webpack config there is a huge difference. Loading svg like this ensure that developers using SDK will not have to configure another loader to load SVGs.

packages/react-native-views/src/components/CameraView/index.jsx
natsukiai: This is not related to this PR. Maybe create a question issue ?

packages/react-native-views/src/components/CameraView/hooks/useFakeActivity.js
natsukiai: This is a correct delay to see something is working, but if it's working too fast, the user will not see a flashing indicator.
Useful when we need to async check device features.

> 1 second: Limit for users feeling that they are freely navigating the command space without having to unduly wait for the computer. A delay of 0.2–1.0 seconds does mean that users notice the delay and thus feel the computer is "working" on the command, as opposed to having the command be a direct effect of the users' actions. Example: If sorting a table according to the selected column can't be done in 0.1 seconds, it certainly has to be done in 1 second, or users will feel that the UI is sluggish and will lose the sense of "flow" in performing their task. For delays of more than 1 second, indicate to the user that the computer is working on the problem, for example by changing the shape of the cursor.

packages/react-native-views/src/components/CameraView/hooks/useUI.js
natsukiai: I added jsdoc comments.

packages/react-native-views/src/components/CameraView/hooks/useFakeActivity.js
rb46546741: OK. I would still give it a name ;)

natsukiai: Done! I've called it `MIN_LOADING_TIME`

packages/react-native-views/src/components/TutorialView/Illustration.jsx
rb46546741: Ok. If it's the same file then would it be possible to make it available as a string and then parse it differently depending on if it's native or not (rather than having it twice)

packages/react-native-views/src/components/CameraView/index.jsx
rb46546741: :sad_panda:

PR #94, opened 2021-11-15
Feature/fe 179/enhance mobile web app

BenHakimIlyass:


natsukiai (commented):
Files reviewed, great job! I'm testing it right away on every devices.

BenHakimIlyass:
The camera view might be a bit smaller on large mobile screens (only if the device is in portrait orientation), but I think it's okay since we inform the user to rotate to landscape for a better experience.

natsukiai:
I have some layout issues while testing.

First on Android Expo App:
Screenshot_20211116-101554

Second on Android web App:
Screenshot_20211116-101802

BenHakimIlyass:
Hmmm, I know why it happens.. will fix it right away 🐛 🐛 .

natsukiai (changes requested):
Fix layout issues. :-)

natsukiai:
> Hmmm, I know why it happens.. will fix it right away 🐛 🐛 .

You're a champ!

BenHakimIlyass:
## Mobile browser layout issue
The first issue is fixed in the latest commit, but the second one is not a bug, perhaps I made it like that on purpose:

When we're on Portrait orientation, and due to the browser address bar provided by IOS we can't get the useful width of the view port(known issue in the community), which gives us the ability to set the right width to the mask and camera as well:

image

And if we try to use the css `100vh` value or Javascript `window.innerHeight` both solutions will give us the whole height including the address bar height, which will result like following:

image

So that's why I made the mask and the camera smaller only on mobile browser and Portrait orientation.

natsukiai (approved):

PR #97, opened 2021-11-16
Add image to inspection on capture

franklinkemta:


franklinkemta:
Actually am getting a 500 internal server error from **/insections/{{inspectionId}}/images** using bot app and postman

PR #98, opened 2021-11-17
Creating a landing page

BenHakimIlyass:


src/store/slices/auth.js
```diff
@@ -19,4 +17,5 @@ export const authSlice = createSlice({
   reducers: createReducers(initialState),
 });
 
+export const { reset, set, update } = authSlice.actions;
```
BenHakimIlyass, line 20: @natsukiai is this the right way to use the actions? or we can access them without export/import?

PR #110, opened 2021-11-24
Feature/fe 115/list damage on vehicle

BenHakimIlyass:
### What's done

- The damaged parts data got passed to the vehicle, and displaying the damage.
- Fixed some incorrect ids from the vehicle SVG file.

### Blocking

- There are some parts id, doesn't exist in our vehicle illustration or are in different spell (e.g we get `rockerPanel` from the backend but we have `rockerPanelLeft` and `rockerPanelRight` in our vehicle SVG, `logo` doesn't exist in our illustration.
- If there is no damages/parts yet in an inspection and we try to go to `damageLibrary` we'll get this error:
image


franklinkemta:
> * rockerPanelRight

PR #124, opened 2021-12-03
Feature/fe 206/damage validation button

BenHakimIlyass:


BenHakimIlyass:
💡 I think we should show the task status somewhere in the damages screen.

BenHakimIlyass:
@natsukiai I don't think we need to have the validate button in the inspection read screen, because every task should validate itself (e.g Inside damages screen, we do validate for the current damages task only), and that's what I did already.

src/screens/Damages/index.jsx
```diff
@@ -47,8 +51,78 @@ const styles = StyleSheet.create({
...
+          Are you sure?
+        </Dialog.Title>
+
+        <Dialog.Content>
+          <Paragraph style={styles.dialogContent}>
+            Please confirm your damage validation
+          </Paragraph>
+        </Dialog.Content>
+
+        <Dialog.Actions>
```
natsukiai, line 100: Can you put the buttons into one CardActions et style the container ? You can see other modifications I've made on dialogs.

natsukiai (changes requested):

PR #141, opened 2021-12-17
Feature/fe 219/manipulate damages create

franklinkemta:
- Update create damage and damagesViews, 
- Fixed DamagePolygon for newly created damage views
- Some fixtutes on damage read

TODO: specify if we need to add polygons to newly created views and how


natsukiai (approved):
This is good work

PR #143, opened 2021-12-20
[FE-227] Add a dialog to save image on upload error

alexandre-em:
Add save image option for web version


natsukiai (changes requested):
This is not what I was asking for.
The feature was:
1. There is an uploading error
2. We alert the user
3. We stop making unnecessary calls
4. When the process is finished, we save pictures on device

PR #146, opened 2021-12-23
[FE-227] Add upload retry icon

alexandre-em:


natsukiai (changes requested):
The refresh button always show, even if the image was uploaded with success. I think it comes from the fact we had it to "notUploadedState" before we get an error.

natsukiai (approved):

PR #201, opened 2022-02-21
Get rid of the annoying useless warnings

BenHakimIlyass:
ALL CONTRIBUTIONS WILL BE REVIEWED BY AT LEAST ONE OF OUR FRONTEND TEAM MEMBERS.

## Overall

Sometimes we get some console errors/warnings that we can't fix as they are coming from a package somehow, e,g **react-navigation** or **react-native-paper** are throwing `EventEmitter.addListener`.

## Technical description

Added a function config which will remove all the given errors/warnings as param, from the console one web, and will remove the also the annoying ones from the native `LogBox`.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -10,6 +10,9 @@ import { Provider as PaperProvider, DefaultTheme } from 'react-native-paper';
 import Navigation from 'Navigation';
 
 import 'config/corejs';
+import ignoreLogs from 'config/ignoreLogs';
+
+ignoreLogs(['EventEmitter.addListener', 'Require cycle'], 'error');
 
 const theme = {
   ...DefaultTheme,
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -10,6 +10,9 @@ import { Provider as PaperProvider, DefaultTheme } from 'react-native-paper';
 import Navigation from 'Navigation';
 
 import 'config/corejs';
+import ignoreLogs from 'config/ignoreLogs';
+
+ignoreLogs(['EventEmitter.addListener', 'Require cycle'], 'error');
```
natsukiai, line 15: I'm ok to ignore `'EventEmitter.addListener'`, but `'Require cycle'` is important. We should add an issue to fix the require circle. I hope that with the SDK 2.0 we will get rid of Require circles.
https://stackoverflow.com/a/10872944

natsukiai (changes requested):

natsukiai (approved):

PR #203, opened 2022-02-22
Upgrade corejs and global SDK to v2.0

natsukiai:
localhost_19006_(iPhone SE) (3)


natsukiai:
I 've requested your review, but it's just to make sure the minimum is working. Don't go deep in the files, you can't obviously. ^^

src/screens/InspectionCreate/index.js
```diff
@@ -1,60 +1,120 @@
...
 
-import useScreen from 'screens/InspectionCreate/useScreen';
+import * as names from 'screens/names';
 
-import { Capture, Controls, Constants, useUploads } from '@monkvision/camera';
+import useAuth from 'hooks/useAuth';
+import useSignIn from 'hooks/useSignIn';
+import useCreateInspection, { TASKS_BY_MOD } from './useCreateInspection';
+
+export const SIGHTS_IDS_BY_MOD = {
```
BenHakimIlyass, line 15: Is this `SIGHTS_IDS_BY_MOD` or `SIGHTS_IDS_BY_MODE`?
Same question inside useCreateInspection.

packages/toolkit/src/hooks/useRequest/index.js
```diff
@@ -0,0 +1,74 @@
...
+    return null;
+  }, [catchLoadingError, onRequestSuccess, request, setSuccess]);
+
+  useEffect(() => {
+    if (typeof request === 'function' && shouldFetch(requestState)) {
+      startRequest();
+    }
+  }, [request, requestState, shouldFetch, startRequest]);
+
+  return { state: requestState, start: request, setters: requestSetState };
```
BenHakimIlyass, line 73: I think we should return `startRequest` as `start` and not `request` as `startRequest` because it's not going to run the `onRequestSuccess` and `catchLoadingError` callbacks (I need this change for commando).

services/inspection-report/src/views/App/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+export const ROUTE_PATHS = {
+  home: '/',
+  inspections: '/inspections',
+};
+
+// Ajoutez ça dans votre fichier de composant
+require('react-dom');
+window.React2 = require('react');
+console.log(window.React1 === window.React2);
```
BenHakimIlyass, line 22: I believe this was related to React version mismatching tests.. to be removed.

BenHakimIlyass:
Inside the capture component we have some props like `onCaptureTourStart`, `onComplianceCheckFinish`… that goes actually to an `useEffect` which takes these callbacks as **dependencies**, the issue here is that `useEffect` will be triggered each time the callbacks changes their values.

Should we remove those callbacks from the useEffects dependencies?

src/screens/InspectionCreate/index.js
natsukiai: Naming is hard. :-( Is mode even a good term for that ?

natsukiai:
> Inside the capture component we have some props like `onCaptureTourStart`, `onComplianceCheckFinish`… that goes actually to an `useEffect` which takes these callbacks as **dependencies**, the issue here is that `useEffect` will be triggered each time the callbacks changes their values.
> 
> Should we remove those callbacks from the useEffects dependencies?

I'm not a fan of props callbacks in big components like this. I prefer to use a state and an API considering the size of it.
We could make three callback props according to the state `onUploadsChange` `onComplianceChange` and `onSightsChange`.

BenHakimIlyass:
Yes I like the idea, and I believe it would be more free and flexible to play with, but I'm not sure if it will be as clear for a client developer who doesn't know how things works behind the scenes.

Anyway I'm with 👍🏼 , but we shall keep an eye on the developer experience.

PR #240, opened 2022-04-20
Improve screen orientation for landscape mode

natsukiai:
We remove the portrait orientation blocker and try a workaround for the orientation.
However, I have know an issue with the camera permissions on android web-app.
If you have info about it let me know.


packages/camera/src/components/Layout/index.js
```diff
@@ -3,10 +3,13 @@ import PropTypes from 'prop-types';
...
-import getOS from '../../utils/getOS';
-import useOrientation from '../../hooks/useOrientation';
 import useMobileBrowserConfig from '../../hooks/useMobileBrowserConfig';
-import PortraitOrientationBlocker from './PortraitOrientationBlocker';
+import useOrientation from '../../hooks/useOrientation';
+
+if (Platform.OS === 'web') {
+  // eslint-disable-next-line global-require
+  require('./layout.css');
+}
```
BenHakimIlyass, line 12: The orientation works great on web, but I couldn't run the app on native due to the following error.

IMG_2038

natsukiai: Yeah of course, that was dumb of me. I need to use RN Stylesheet. ^^

PR #242, opened 2022-04-26
corejs typescript migration 

souyahia-monk:


packages/corejs/src/damages/index.ts
```diff
@@ -38,7 +38,7 @@ export const TYPES = {
  * @param {string} data.partType
  * @param {Object} requestConfig
  */
-export const createOne = async ({ inspectionId, data, ...requestConfig }) => {
+export const createOne = async ({ inspectionId, data, ...requestConfig }: AxiosRequestConfig & { inspectionId: any }) => {
```
natsukiai, line 41: Shouldn't be inspectionId of type string ?

packages/corejs/src/inspections/index.ts
```diff
@@ -75,25 +75,30 @@ export const getMany = async ({ params, ...requestConfig }) => {
...
+    const entity = { ...data, [idAttribute]: id, createdAt };
+
+    return ({
+      axiosResponse,
+      [idAttribute]: id,
+      ...normalize(entity, schema),
+    });
+  } catch(err) {
+    console.log('Error found !!');
+    console.log(err);
```
natsukiai, line 99: We should let the developer catch his.her error him.herself or `throw err;` in our own catch.

packages/corejs/src/tasks/index.ts
```diff
@@ -32,11 +32,11 @@ export const STATUSES = {
  * @param {string} name
  * @param {Object} requestConfig
  */
-export const getOne = async ({ inspectionId, name, ...requestConfig }) => {
+export const getOne = async ({ inspectionId, name, ...requestConfig }: AxiosRequestConfig & { inspectionId: any, name: any }) => {
   const axiosResponse = await axios.request({
     ...config.axiosConfig,
     method: 'get',
-    url: `/inspections/${inspectionId}/task/${name}`,
+    url: `/inspections/${inspectionId}/tasks/${name}`,
```
natsukiai, line 39: Good catch!

packages/corejs/test/examples/damage.example.ts
```diff
@@ -0,0 +1,6 @@
+const damageExample = {
+  damage_type: 'body_crack',
+  part_type: 'ignore'
+}
+
+export default damageExample;
```
natsukiai, line 6: You can set keys in camelCase even in examples because we are mapping deep the keys to transform snake_case to camelCase. I'm not happy about it, but this is what we expected working with JS.

natsukiai (changes requested):

natsukiai (approved):

PR #254, opened 2022-05-03
Feature/camera resolution

BenHakimIlyass:
ALL CONTRIBUTIONS WILL BE REVIEWED BY AT LEAST ONE OF OUR FRONTEND TEAM MEMBERS.

## Technical description

Re-implemented the camera logic, with a fixed resolution `2560x1440`.

The solution is to use a resolution of `2561x1441` in both canvas and video constraints, but when we take the picture we clip the image down to `2560x1440`.

As long as I tested, when we use a canvas (which called `baseCanvas`) and a video constraints of `2560x1440` resolution, the app crashes, maybe because the canvas doesn't use the exact amount of pixels, however as a solution I increased the `baseCanvas` and the video constraints resolution to `2561x1441` and created another canvas called `croppedCanvas` which will be taking picture based on `baseCanvas` (not based on the `video.current`) while holding a resolution of `2560x1440`, as a result it gives file size of around 5mb in `png` and around 200kb in `webp`.

Please the find the tested devices/browsers [here](https://docs.google.com/spreadsheets/d/1y2KWqxRaXMVEIsv7qtOD5WL2E61MIRtBTVPRZFgEzus/edit?usp=sharing).
(Feel free to add your tested device to the sheet).

## Related to

This PR is related to camera component.

## Tests
Steps to reproduce

- Use the camera on different browsers/devices
- Expected: the app shouldn't crash after few images

## Screenshots (optional):


packages/camera/src/components/Camera/hooks/useUserMedia.js
```diff
@@ -0,0 +1,43 @@
...
+      try {
+        const str = await navigator.mediaDevices.getUserMedia(constraints);
+        if (!didCancel) { setStream(str); }
+      } catch (err) {
+        if (!didCancel) { setError(err); }
+      }
+    };
+
+    getUserMedia();
+    // eslint-disable-next-line consistent-return
```
souyahia-monk, line 26: Instead of disabling this eslint error, couldn't we just make the function return `() => {}` on line 12 ?

BenHakimIlyass: Yes, we can for sure.

alexandre-em (approved):
Tested with iphone8, samsung galaxy j7 with chrome, firefox and samsung internet/safari. It's ok

PR #277, opened 2022-05-24
Feature/error catching

souyahia-monk:


packages/toolkit/src/hooks/useError/index.js
```diff
@@ -0,0 +1,8 @@
+import { useCallback } from 'react';
+
+export default function useError() {
+  return useCallback((error) => {
+    console.error('Error detected !!!!!!');
```
alexandre-em, line 5: I don't think we will put the sentry on the sdk, so perhaps we can remove this hook.. what do you think?

natsukiai: We could put Sentry on the SDK with an Opt-in param and a dedicated dsn for customers.

natsukiai (commented):
Just need a console.log() clean up

natsukiai (changes requested):
Fix eslint errors

natsukiai:
@souyahia-monk can we remove or add eslint ignores for the `console.error` making _github_ code scan throw a warning in the file changes ?

PR #309, opened 2022-06-21
feat: Save Auth state

souyahia-monk:


src/components/App.js
```diff
@@ -63,7 +64,7 @@ function App() {
         await Font.loadAsync(customFonts);
         // Artificially delay for two seconds to simulate a slow loading
         // experience. Please remove this if you copy and paste the code!
-        await new Promise((resolve) => { setTimeout(resolve, 2000); });
+        // await new Promise((resolve) => { setTimeout(resolve, 2000); });
```
natsukiai, line 67: Actually, I was liking the first loader saying "Launching the app" ^^ but as the comment says, we can disable it.

src/screens/Landing/SignOut/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <Menu
+      visible={visible}
+      onDismiss={closeMenu}
+      anchor={(
+        <List.Subheader style={styles.subheader}>
+          <IconButton icon="account-circle" onPress={openMenu} size={32} />
+          You are connectedsq
+        </List.Subheader>
```
natsukiai, line 44: Here's a typo ^^

src/screens/Landing/SignOut/index.js
```diff
@@ -0,0 +1,51 @@
...
+      )}
+      style={{ zIndex: 999 }}
```
natsukiai, line 46: Is this z-index mandatory ? I'm really not a fan of absolute z-index.

src/components/App.js
souyahia-monk: Actually, I forgot to comment back this line. I just commented it for faster testing, should I add it back ?

natsukiai: Yes please. :-)

natsukiai (approved):

Issue #319, opened 2022-06-29
Translation of some components of the SDK


natsukiai:
### Discussed in https://github.com/monkvision/monkjs/discussions/318

<div type='discussions-op-text'>

<sup>Originally posted by **Nicolas-Schuhl** June 28, 2022</sup>
**Pain points:**

- Currently the Capture Component is only in English some developers require the possibility to have the capture component in french

**Useful:** 
- https://www.notion.so/monkvision/Update-Toyota-web-app-0137efb0f196479ca94471a2eb3f6c5a
- Existing mappings for parts, damages et vehicule details: [https://docs.google.com/spreadsheets/d/1xesvhBEPBOkOOHYluGgqeEKFn-E-4XDH64BeekmbJxg/edit?usp=sharing](https://docs.google.com/spreadsheets/d/1xesvhBEPBOkOOHYluGgqeEKFn-E-4XDH64BeekmbJxg/edit?usp=sharing)</div>

natsukiai:
https://github.com/monkvision/monkjs/discussions/318#discussioncomment-3046687

PR #329, opened 2022-07-13
Feature #319 : Webapp, Capture and Sights french translation

souyahia-monk:


packages/camera/src/i18n/resources/en.js
```diff
@@ -0,0 +1,70 @@
+const en = {
+  translation: {
+    camera: {
+      permissionDenied: 'Impossible d\'accéder à la caméra : permission refusée.',
```
alexandre-em, line 4: It's in french instead of english and the same for fr.js it is in english instead of french

BenHakimIlyass:
The mods texts on the landing screen are not translated

image

A suggestion for later:
We shall start using icons in the capture buttons instead of texts for better UI.

image

natsukiai (approved):

PR #382, opened 2022-11-24
Prettify the sights json (to help future updates)

artemisart:
Hello, this is a separate PR to avoid a dense diff of 10k lines, I used prettier with default parameters to help minize the diff and review time :) for the next update of the sights json.

## Technical description

- [x] Feature description...

## Tested on

-
- 

## Steps to reproduce

1. 
2. 

## Screenshots (optional):

No screenshots.


packages/sights/package.json
```diff
@@ -31,12 +31,15 @@
     "README.md"
   ],
   "scripts": {
-    "prepare": "yarn run lint && yarn run build",
+    "prepare": "npm run lint && npm run build",
```
souyahia-monk, line 34: why switch to npm here ?

packages/sights/pnpm-lock.yaml
```diff
@@ -0,0 +1,15 @@
+lockfileVersion: 5.4
```
souyahia-monk, line 1: on ne devrait pas avoir de pnpm-lock sur ce package, juste un yarn.lock (la version de yarn devrait déjà être fixée)

packages/sights/package.json
artemisart: to be compatible with other package managers, here I used pnpm, and for `run` is enough and bundled with node so it will run with anything

PR #403, opened 2023-01-18
MN-155 : MonkJs Sights 4.0.0

souyahia-monk:


monkjs-4.0.0/packages/sights/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@monkvision/sights",
```
rohitsharma120582, line 2: Don't have version detail...

rohitsharma120582 (approved):

PR #408, opened 2023-01-27
feat: Remove useless sentry code (MN-182)

deepakglobant:
## Technical description
Removed useless sentry/monitoring code

## Screenshots (optional):

No screenshots.


rohitsharma120582 (commented):
LGTM! However, I spot some pieces which aren't removed instead commented. I think, you can keep only a comment to describe what needs to be done there and remove that code as well.

deepakglobant:
> LGTM! However, I spot some pieces which aren't removed instead commented. I think, you can keep only a comment to describe what needs to be done there and remove that code as well.

Yeah agree. I removed the commented code and add a todo to add new code in next ticket.

souyahia-monk (approved):

PR #414, opened 2023-02-02
Fix stretched image issue | HGreg

rohitsharma120582:
## Description
Fix stretched image issue for sights pic in Android tablet device (like Samsung Galaxy Tab A  Model: SM-T307U)

## Technical description
Recalculating the resolution(dimension) for canvas as per video's aspect ratio when drawing the image:
- If video's aspect ratio is 4:3, then canvas resolution would be either FHD (1920x1080) or QHD (2560x1440) based on selected resolution
- If video's aspect ratio is not 4:3 then going to calculate new resolution as per video's aspect ratio and set to canvas


deepakglobant (approved):
LGTM!

souyahia-monk (approved):

PR #422, opened 2023-02-03
Feature/mn 175/factorize monitor code

deepakglobant:
## Technical description
- Create monitoring context to share functions and sentry config across the applications.
- Initialise monitoring through context.
- No need to pass config for sentry in each an every places.
- Created errorMessage and measureMonitoring functionality.
- Added custom hooks (useMonitoring) that will provide all the monitoring functions to use in each an every packages.


rohitsharma120582 (approved):
LGTM!

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { Integrations } from '@sentry/tracing';
+import { Platform } from 'react-native';
+import * as Sentry from 'sentry-expo';
+import * as SentryR from '@sentry/react';
+import Constants from 'expo-constants';
+import PropTypes from 'prop-types';
+
+import { MonitoringContextType } from './types';
+
+const tracingOrigins = ['localhost', 'cna.dev.monk.ai', 'cna.staging.monk.ai', 'cna.preview.monk.ai', 'cna.monk.ai'];
```
souyahia-monk, line 13: What do the tracing origins do exactly ? Because I don't think we should hard code them like that. We should allow users to pass them as configuration parameters for the useMonitoring hook. There are A LOT of other URLs in which the SDK is used. Just in Monk alone there are at least 20 different URLs where the SDK is used, and this is without taking the Clients into account.

By the way there are some mistakes in the URLs here because for instance the stagng URL is the following : `cna-staging.dev.monk.ai`

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+const MonitoringConstants = {
+  type: {
```
souyahia-monk, line 15: Use a TypeScript enum instead of a JS Object

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    UPLOAD: 'upload', // logs linked to the upload
+    CAMERA: 'camera', // logs linked to the camera
+    FUNC: 'func', // logs linked to a function
+    APP: 'app', // logs linked to the application
+    HTTP: 'http', // logs linked to the api
+  },
+  operation: {
```
souyahia-monk, line 22: Same as above, use a TypeScript enum

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    USER_UPLOAD_CENTER_TIME: 'user-upload-center-time',
+    USER_ACTION: 'user-action',
+    RESPONSE_TIME: 'response-time',
+    FUNC: 'func',
+  },
+};
+
+const Tracing = Integrations.BrowserTracing;
+const setTag = (key: string, value: Primitive): void => SentryR.setTag(key, value);
+const setUser = (id: string) => SentryR.setUser({ id });
```
souyahia-monk, line 35: use function syntax instead of arrow syntax

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+const MonitoringContext = createContext<MonitoringContextType | null>(null);
+
+function MonitoringProvider({ children, config }) {
```
souyahia-monk, line 38: provide type definition for parameters and return type

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  useEffect(() => {
+    Sentry.init({
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      dsn: (config.dns || Constants.manifest.extra.SENTRY_DSN) as string,
+      environment: Constants.manifest.extra.ENV as string,
+      debug: Constants.manifest.extra.ENV !== 'production',
```
souyahia-monk, line 44: You CAN't use Constants.manifest in the CoreJs package. Constants.manifest is only available in Expo projects, and we DO NOT have expo installed in the CoreJs package.

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+      enableAutoSessionTracking: true,
+      enableInExpoDevelopment: true,
+      sessionTrackingIntervalMillis: 10000,
+      tracesSampleRate: Constants.manifest.extra.ENV !== 'production' ? 1.0 : 0.2,
```
souyahia-monk, line 48: All of this configuration object should not be a magic value and should be defined as a const somewhere

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    } as Sentry.SentryExpoNativeOptions);
+  }, []);
+
+  /**
+   * @param error {Error} - Caught error to send to Sentry.io
+   * @param type {string} - tag of the error's type
+   * @param extras {Record} - Useful information that can be sent (request's body for example)
+   * @param additionalTags - (Optional) Additional tags to add to the error
+   */
+  const errorHandler = (error: Error | string): string => {
```
souyahia-monk, line 67: une function syntax instead of arrow syntax

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+        }),
+      ],
+    } as Sentry.SentryExpoNativeOptions);
+  }, []);
+
+  /**
+   * @param error {Error} - Caught error to send to Sentry.io
+   * @param type {string} - tag of the error's type
+   * @param extras {Record} - Useful information that can be sent (request's body for example)
+   * @param additionalTags - (Optional) Additional tags to add to the error
```
souyahia-monk, line 65: You are declaring other parameters such as type, extras and additional tags but I don't see them in the function signature. Either remove them from the JSDoc or add them in the function signature and use them

souyahia-monk, line 65: remove the (Optional) mention and instead put the parameter inside square brackets [] (please checkout this link for future reference : https://jsdoc.app/tags-param.html#optional-parameters-and-default-values)

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+          native: [],
+        })),
+        new Tracing({
+          tracingOrigins,
+        }),
+      ],
+    } as Sentry.SentryExpoNativeOptions);
+  }, []);
+
+  /**
```
souyahia-monk, line 61: please add a small description of what this function do

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+   * @param error {Error} - Caught error to send to Sentry.io
+   * @param type {string} - tag of the error's type
+   * @param extras {Record} - Useful information that can be sent (request's body for example)
```
souyahia-monk, line 64: if you use the Record type, please provide type definitions for the generic args of the Record type

packages/corejs/src/monitoring/index.tsx
souyahia-monk, line 65: please provide type definitions for this argument

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+   * @param additionalTags - (Optional) Additional tags to add to the error
+   */
+  const errorHandler = (error: Error | string): string => {
+    if (!Sentry || (!Sentry?.Browser && !Sentry?.Native)) { return null; }
```
souyahia-monk, line 68: The function you declared is supposed to be returning a string element, but you are returning a null element here. Either change the return type to `string | null` or return `''` in this case;

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    return Platform.select({
+      web: Sentry.Browser,
+      native: Sentry.Browser,
+    }).captureException(error);
+  };
+
+  /**
```
souyahia-monk, line 76: Please provide a small description of what this function do

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+   * @param name {string} - Name of transaction
+   * @param operation {string} - Operation of transaction to be performed
+   * @param data {{[key: string]: number | string}} - Data to be added on transaction
```
souyahia-monk, line 79: I noticed that the data parameter is optional. Please make it optional in the JSDoc : https://jsdoc.app/tags-param.html#optional-parameters-and-default-values

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+   */
+  const measurePerformance = (name: string, operation: string, data: { [key: string]: number | string } | null) => {
```
souyahia-monk, line 81: Please use the function syntax instead of the arrow syntax. Also : make the data param optional instead of making it a union with null. Also, please specify a return type

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    const transaction = Platform.select({
+      web: Sentry.Browser,
+      native: Sentry.Browser,
+    }).startTransaction({ name, data });
```
souyahia-monk, line 85: We are using this Platform.select code a lot, maybe we should factorize it somewhere ?

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  );
+
+  return (
+    <MonitoringContext.Provider value={monitoringContextValue}>
+      {children}
+    </MonitoringContext.Provider>
+  );
+}
+
+const useMonitoring = () => useContext(MonitoringContext);
```
souyahia-monk, line 106: use the function syntax instead of the arrow syntax

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export {
+  MonitoringConstants,
+  MonitoringContext,
+  setTag,
+  setUser,
+  Tracing,
+  MonitoringProvider,
+  useMonitoring,
+};
```
souyahia-monk, line 116: can you add the export keyword directly to the elements exported instead of having an export object at the end ? I think it's clearer

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  config: PropTypes.shape({
+    dns: PropTypes.string,
+  }),
+};
+
+MonitoringProvider.defaultProps = {
+  config: {
+    dns: '',
+  },
+};
```
souyahia-monk, line 128: Since we are in TypeScript I don't think we need PropTypes, we can actually use TypeScript type defs which is better

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Monitoring context interface
+ */
+export interface MonitoringContext {
```
souyahia-monk, line 4: where is this interface used ?

packages/corejs/src/monitoring/index.tsx
deepakglobant: We are getting value for Constants.manifest in CoreJs package.

packages/corejs/src/monitoring/index.tsx
deepakglobant: ah yeah. got it.

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Monitoring context which will create wrapper for monitoring functionality.
+*/
+export const MonitoringContext = createContext<MonitoringContextType | null>(null);
+
+/**
+ * Monitoring wrapper used to abstract Sentry functionality.
+ *
+ * @param {object} children - Provider children components that will be shared inside context provider.
+ * @param {MonitoringConfig} config - Configuration for sentry to override default configuration.
```
souyahia-monk, line 20: The params provided here are actually members of an object in the function signature. So in the JSDoc they should not be displayed as function params but as object properties

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+      ],
+    });
+  }, []);
+
+  /**
+   * Updates user context information for future events.
+   *
+   * @param id {string} set user for in sentry
+  */
+  const setMonitoringUser = useCallback((id: string): void => {
```
souyahia-monk, line 40: Why are you specifying the `: void` return type ? You can remove this

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * Set key:value that will be sent as tags data with the event.
+   *
+   * Can also be used to unset a tag, by passing `undefined`.
+   *
+   * @param key String key of tag
+   * @param value Value of tag
+  */
+  const setMonitoringTag = useCallback((key: string, value: Primitive): void => {
```
souyahia-monk, line 52: You can remove the `: void` return type

packages/corejs/src/monitoring/index.tsx
souyahia-monk: @deepakglobant  you still need to put the parameter as optional in the JSDoc (between square brackets).

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * Measure the performance of application based on functionality and operation based on it.
+   * Return type of the function is the IIFE, which will helps to close the transaction and complete the measurement.
+   *
+   * @param name {string} - Name of transaction
+   * @param operation {string} - Operation of transaction to be performed
+   * @param data {{[key: string]: number | string}} - Data to be added on transaction
+   */
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  const measurePerformance = useCallback((name: string, op: string, data?: { [key: string]: number | string }): Function => {
```
souyahia-monk, line 78: The @typescript-eslint/ban-types rule is there for a reason ! The `Function` type should NOT be used. Please remove the eslint comment and replace the `Function` type with a type definition for the callback returned

packages/corejs/src/monitoring/index.tsx
souyahia-monk: @deepakglobant Also can you specify the return type of the function in the JSDoc please

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }, []);
+
+  /**
+   * Set key:value that will be sent as tags data with the event.
+   *
+   * Can also be used to unset a tag, by passing `undefined`.
+   *
+   * @param key String key of tag
+   * @param value Value of tag
+  */
```
souyahia-monk, line 51: please specify the return type of the function in the JSDoc

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+    };
+  }, []);
+
+  /**
+   * Set the custom measurement on particular transaction
+   *
+   * @param transactionName Name of the transaction
+   * @param name Name of the measurement
+   * @param value Value of the measurement
+   * @param unit Unit of the measurement. (Defaults to an empty string)
```
souyahia-monk, line 100: make the unit param optional in the JSDoc

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,92 @@
...
+   * all traces)
+   *
+   * Tracing is enabled if either this or `tracesSampler` is defined. If both are defined, `tracesSampleRate` is
+   * ignored.
+  */
+  tracesSampleRate: number;
+  /**
+   * Array of all the origin to browser trace.
+  */
+  tracingOrigins: Array<string>;
```
souyahia-monk, line 30: replace `Array<string>` by `string[]`

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/**
+ * Monitoring context interface
+*/
+export interface MonitoringContextType {
```
souyahia-monk, line 36: Remove the `Type` at the end of the type name

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Store the error in the monitoring application.
+  */
+  errorHandler: (error: Error | string) => string;
+
+  /**
+   * Start Measure Performance
+  */
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  measurePerformance: (name: string, op: string, data: { [key: string]: number | string } | null) => Function;
```
souyahia-monk, line 51: Same comment as before : remove the eslint comment and replace the `Function` type with actual type definitions for the returned callback

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Set custom measurement value
+  */
+  setMeasurement: (transactionName: string, name: string, value: number, unit: string) => void;
+}
+
+/**
+ * Monitoring configuration interface
+*/
+export interface MonitoringConfigType {
```
souyahia-monk, line 62: Remove the `Type` at the end of the type name

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Provider children components that will be shared inside context provider.
+  */
+  children: object;
```
souyahia-monk, line 66: Can we use anything else other than object ? Maybe unknown ? It is a bad practice to use the object type

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  FUNC = 'func',
+}
+
+export enum MonitoringOperations {
+  UPLOAD = 'upload',
+  CAMERA = 'camera',
+  FUNC = 'func',
+  APP = 'app',
+  HTTP = 'http',
+}
```
souyahia-monk, line 90: where are these enums used ? If they are not used we can remove them

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+*/
+export function MonitoringProvider({ children, config }: MonitoringConfigType) {
+  useEffect(() => {
+    Sentry.init({
+      dsn: (config.dsn || Constants.manifest.extra.SENTRY_DSN) as string,
+      environment: (config.environment || Constants.manifest.extra.ENV) as string,
+      debug: config.debug || Constants.manifest.extra.ENV !== 'production',
+      tracesSampleRate: config.tracesSampleRate || 0.5,
+      integrations: [
+        new BrowserTracing({ tracePropagationTargets: config.tracingOrigins }),
```
souyahia-monk, line 30: please don't use Sentry constants in the corejs package. Only use the configuration object given as parameter. We can't use expo in corejs

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+*/
+export const MonitoringContext = createContext<MonitoringContextType | null>(null);
+
+/**
+ * Monitoring wrapper used to abstract Sentry functionality.
+ *
+ * @param {object} children - Provider children components that will be shared inside context provider.
+ * @param {MonitoringConfig} config - Configuration for sentry to override default configuration.
+*/
+export function MonitoringProvider({ children, config }: MonitoringConfigType) {
```
souyahia-monk, line 22: add a return type to this

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+   * @param name Name of the measurement
+   * @param value Value of the measurement
+   * @param unit Unit of the measurement. (Defaults to an empty string)
+   */
+  const setMeasurement = useCallback((transactionName: string, name: string, value: number, unit?: string): void => {
+    const transaction = Sentry.startTransaction({ name: transactionName, op: name });
+
+    setTimeout(() => {
+      transaction.setMeasurement(name, value, unit);
+      transaction.setMeasurement('frames_total', value, unit);
```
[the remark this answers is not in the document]
deepakglobant: I added the same just to get the same records in default columns. As we are not getting custom columns in table and graphs.

packages/corejs/src/monitoring/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Measure the performance of application based on functionality and operation based on it.
+   * Return type of the function is the IIFE, which will helps to close the transaction and complete the measurement.
+   *
+   * @param name {string} - Name of transaction
+   * @param operation {string} - Operation of transaction to be performed
+   * @param [data] {{[key: string]: number | string}} - Data to be added on transaction
+   * @returns {IMeasurePerformance} - Which will helps to close the transaction and complete the measurement.
+  */
+  const measurePerformance = useCallback((name: string, op: string, data?: { [key: string]: number | string }): IMeasurePerformance => {
```
souyahia-monk, line 80: replace `IMeasurePerformance` with `() => void`

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,77 @@
...
+  tracesSampleRate: number;
+  /**
+   * Array of all the origin to browser trace.
+  */
+  tracingOrigins: string[];
+}
+
+export interface IMeasurePerformance {
+  (): void
+}
```
souyahia-monk, line 35: remove this useless type

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Monitoring context interface
+*/
+export interface IMonitoringContext {
```
souyahia-monk, line 40: Remove the I at the start of the type name. Call it just "MonitoringContext"

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Store the error in the monitoring application.
+  */
+  errorHandler: (error: Error | string) => string;
+
+  /**
+   * Start Measure Performance
+  */
+  measurePerformance: (name: string, op: string, data: { [key: string]: number | string } | null) => IMeasurePerformance;
```
souyahia-monk, line 54: replace `IMeasurePerformance` with `() => void`

packages/corejs/src/monitoring/types.ts
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * Set custom measurement value
+  */
+  setMeasurement: (transactionName: string, name: string, value: number, unit: string) => void;
+}
+
+/**
+ * Monitoring configuration interface
+*/
+export interface IMonitoringProvider {
```
souyahia-monk, line 65: Remove the I at the start of the type name, call it just "MonitoringProviderProps" or something like that

packages/corejs/src/monitoring/index.tsx
deepakglobant: Done

packages/corejs/src/monitoring/types.ts
deepakglobant: Yup. Got it and updated

souyahia-monk (approved):

PR #436, opened 2023-02-23
feat: replace settings button with full screen(MN-197)

deepakglobant:
- [x] Feature description...
Replace settings button with full screen button
Added conditional based settings button which will toggle settings and full screen button.

### DIFF
--- a/packages/camera/src/components/Controls/FullScreenButton.js
+++ b/packages/camera/src/components/Controls/FullScreenButton.js
@@ -3,7 +3,7 @@ import React from 'react';
 import { Text } from 'react-native';
 import { MaterialCommunityIcons } from '@expo/vector-icons';
 
-export default function FullScreenButton({ label }) {
+export default function FullScreenButton({ isInFullScreenMode }) {
   return (
     <Text
       style={{
@@ -15,22 +15,15 @@ export default function FullScreenButton({ label }) {
         textTransform: 'uppercase',
       }}
     >
-      {
-        label === 'fullScreen'
-        && <MaterialCommunityIcons name="fullscreen" size={24} />
-      }
-      {
-        label === 'exitFullScreen'
-        && <MaterialCommunityIcons name="fullscreen-exit" size={24} />
-      }
+      <MaterialCommunityIcons name={isInFullScreenMode ? 'fullscreen-exit' : 'fullscreen'} size={24} />
     </Text>
   );
 }
 
 FullScreenButton.propTypes = {
-  label: PropTypes.string,
+  isInFullScreenMode: PropTypes.bool,
 };
 
 FullScreenButton.defaultProps = {
-  label: 'fullScreen',
+  isInFullScreenMode: true,
 };
--- a/packages/camera/src/components/Controls/FullScreenButton.js
+++ b/packages/camera/src/components/Controls/FullScreenButton.js
@@ -1,10 +1,10 @@
-import React from 'react';
+import PropTypes from 'prop-types';
 import { useTranslation } from 'react-i18next';
+import React from 'react';
 import { Text } from 'react-native';
 
-export default function FullScreenButton() {
+export default function FullScreenButton({ label }) {
   const { t } = useTranslation();
-
   return (
     <Text
       style={{
@@ -16,7 +16,11 @@ export default function FullScreenButton() {
         textTransform: 'uppercase',
       }}
     >
-      {t('controls.fullScreen')}
+      {t(label)}
     </Text>
   );
 }
+
+FullScreenButton.propTypes = {
+  label: PropTypes.string.isRequired,
+};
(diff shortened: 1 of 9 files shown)


src/screens/InspectionCapture/index.js
```diff
@@ -146,8 +149,22 @@ export default function InspectionCapture() {
...
+    utils.log(['[Click] Camera setting: ', 'Fullscreen ', !isFullscreen]);
+    requestFullscreen();
+  }, [isFullscreen]);
+
   const controls = [
-    { disabled: cameraLoading, ...Controls.SettingsButtonProps, onPress: openSettings },
+    !isSettingEnabled && {
+      disabled: cameraLoading,
+      ...Controls[isFullscreen ? 'ExitFullscreenButtonProps' : 'FullscreenButtonProps'],
+      onPress: handleFullScreen,
```
rohitsharma120582, line 161: No need to declare and call `handleFullScreen` function. You can directly call `requestFullscreen` function here.

rohitsharma120582: Also you should make a condition to render `<Settings ref={settingsRef} settings={settings} />` component as per `isSettingEnabled` value...

packages/camera/src/components/Controls/FullScreenButton.js
```diff
@@ -16,7 +16,11 @@ export default function FullScreenButton() {
         textTransform: 'uppercase',
       }}
     >
-      {t('controls.fullScreen')}
+      {t(label)}
     </Text>
   );
 }
+
+FullScreenButton.propTypes = {
```
rohitsharma120582, line 24: Plz also add `FullScreenButton.defaultProps` for default value. It can be empty string here.

souyahia-monk:
Hey @deepakglobant so I took a look at your code, and I think there was a bit of a miss understanding on the feature we need for this ticket. We don't want to make the Settings button _optional_, but we want to _remove_ it. The goal of this feature is to completely remove the settings button and replace it with just a fullscreen button.

Sorry, this ticket was not very properly specified. I have updated the ticket with more precise information.

packages/camera/src/components/Controls/FullScreenButton.js
```diff
@@ -1,10 +1,9 @@
+import PropTypes from 'prop-types';
 import React from 'react';
-import { useTranslation } from 'react-i18next';
 import { Text } from 'react-native';
+import { MaterialCommunityIcons } from '@expo/vector-icons';
```
souyahia-monk, line 4: If you use `@expo/vector-icons` you need to add it to the dependencies of the package.

packages/camera/src/components/Controls/FullScreenButton.js
```diff
@@ -1,10 +1,9 @@
...
 
-export default function FullScreenButton() {
-  const { t } = useTranslation();
-
+export default function FullScreenButton({ label }) {
```
souyahia-monk, line 6: instead of a `label` prop i think it's better to use a boolean prop called `isInFullScreenMode`

packages/camera/src/components/Controls/index.js
```diff
@@ -363,6 +363,11 @@ Controls.FullscreenButtonProps = {
   },
 };
 
+Controls.ExitFullscreenButtonProps = {
+  ...Controls.FullscreenButtonProps,
+  children: <FullScreenButton label="exitFullScreen" />,
+};
+
```
souyahia-monk, line 370: instead of having 2 sets of props (`Controls.FullscreenButtonProps` and `Controls.ExitFullscreenButtonProps`) I think it would be better to have only one prop and pass a boolean parameter indicating if the app is in full screen mode

packages/camera/src/components/Controls/index.js
```diff
@@ -341,7 +341,7 @@ Controls.AddDamageButtonProps = {
 Controls.FullscreenButtonProps = {
```
souyahia-monk, line 341: also please add a check in the Controls component to check if we are using Safari. If we are using Safari, we should add `visibility: 'hidden'` to the button with id `'full-screen'`

rohitsharma120582 (approved):

packages/camera/src/i18n/resources/fr.js
```diff
@@ -10,6 +10,7 @@ const fr = {
     controls: {
       takePicture: 'Prendre La Photo',
       fullScreen: 'Plein Écran',
+      exitFullScreen: 'Sortie Plein Écran',
```
souyahia-monk, line 13: Replace `Sortie` by `Quitter`

src/screens/InspectionCapture/index.js
```diff
@@ -164,12 +166,12 @@ export default function InspectionCapture() {
...
-  const settingsRef = useRef();
-  const openSettings = settingsRef.current?.open;
-
   const controls = [
-    { disabled: cameraLoading, ...Controls.SettingsButtonProps, onPress: openSettings },
+    !isSafari && {
+      disabled: cameraLoading,
+      ...Controls.getFullScreenButtonProps(isFullscreen),
+      onPress: requestFullscreen,
+    },
```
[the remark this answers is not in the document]
deepakglobant: Done.

souyahia-monk (approved):

PR #445, opened 2023-03-08
feat: generate tomm metrics manually (MN-221)

deepakglobant:


souyahia-monk:
Hello, a quick comment to summarize what we discussed this morning :

- Place the script inside directory called `scripts` at the root of the repository
- Create a package.json inside this directory ()you can call the project `@monkvision/scripts` or something like that, don't forget to declare the project as private in the package.json)
- Add the required dependencies for the scripts and everything inside this npm project, not the root 
- The NPM script used to run the script should also be declared in this project
- Add a .env file in the scripts directory, and add it to gitignore
- Add the "dotenv" dependency to the scripts project, and at the start of the script, use dotenv to load the .env file containing the variables (you can check the .env documentation on how to do it, it's very easy)
 - We need to be able to run the script without any parameters, all config should be in the .env file
 - We should be able to configure the following properties in the .env file :
   - The sentry token
   - The environment to fetch (staging / preview etc...)
   - The projects we want to fetch (capture-app, toyota-develop, toyota-prerestitution etc...)

souyahia-monk:
Hey @deepakglobant , good work on this. Could you also add a very small README.md in the scripts directory to explain how to run it (listing the config env vars etc...). Thanks a lot

rohitsharma120582 (approved):

scripts/Readme.md
```diff
@@ -0,0 +1,15 @@
...
+``` yarn
+yarn run extract-metrics
+```
+
+We need a environment (.env) file at the same directory with below keys
+
+| **Keys**                 | **Description**                                                                              |
+|--------------------------|----------------------------------------------------------------------------------------------|
+| `TOKEN`                  | Sentry auth_token to access the api's                                                        |
+| `PROJECTS`               | List of all projects if user wants to filter else pass empty value (Ex. "")                  |
```
souyahia-monk, line 14: Is this a comma-separated list ? It could be nice to specify in the doc I think

deepakglobant: Done

souyahia-monk (approved):

PR #476, opened 2023-04-18
feat: added gallery component (MN-253)

deepakglobant:
Added Gallery component


packages/inspection-report/src/assets/camera.svg
rohitsharma120582, line 1: Why do we have this SVG file as we are using MaterialIcon to show the icons?

deepakglobant: Agree and removed.

packages/inspection-report/src/components/Gallery.js
rohitsharma120582, line 1: First of all, plz get the idea about the source of gallery data whether it's going to come from backend database or manage at client side directly. Also, if you keep multiple labels for different languages, then implementation the logic to change the label of photos according to selected language.

packages/inspection-report/src/components/Gallery.js
```diff
@@ -0,0 +1,180 @@
...
+      setVisibleImageInFullScreen(true);
+    }
+  }, [photo, visibleImageInFullScreen]);
+
+  return (
+    <View style={styles.container}>
+      {
+        pictureList.map((image, index) => (
+          // eslint-disable-next-line react/no-array-index-key
+          <Thumbnail key={`${image.url}-${index}`} image={image} click={handleOnImageClick} />
```
souyahia-monk, line 135: I think you can remove the index from the key, the image URL should be unique

packages/inspection-report/src/components/Gallery.js
```diff
@@ -0,0 +1,180 @@
...
+      <Modal
+        animationType="slide"
+        transparent
+        visible={visibleImageInFullScreen}
+        onRequestClose={() => {
+          setPhoto('');
+          setVisibleImageInFullScreen(!visibleImageInFullScreen);
+        }}
+      >
+        <view style={{ flex: 1 }}>
```
souyahia-monk, line 148: missing capital letter on View

packages/inspection-report/src/components/Gallery.js
```diff
@@ -0,0 +1,180 @@
...
+    paddingHorizontal: 15,
+    paddingVertical: 5,
+    backgroundColor: '#121212',
+  },
+  header: {
+    fontSize: 20,
+  },
+});
+
+const pictureList = [
```
souyahia-monk, line 32: 1. Create a file named `mock.js` in the `src` directory
2. Place these fake pictures in this file and export them
3. Import the pictures in the DamageReport component
4. Pass them as a prop to the Gallery component

packages/inspection-report/src/components/Gallery.js
```diff
@@ -0,0 +1,180 @@
...
+    url: '',
+    label: {
+      fr: 'Bike',
+      en: 'Bike',
+    },
+  },
+];
+
+export default function Gallery() {
+  const [photo, setPhoto] = useState('');
```
souyahia-monk, line 120: rename variable to `focusedPhoto`

packages/inspection-report/src/components/Gallery.js
```diff
@@ -0,0 +1,180 @@
...
+  const [visibleImageInFullScreen, setVisibleImageInFullScreen] = useState(false);
```
souyahia-monk, line 121: no need for this variable. You can know when the modal must be displayed based on if the photo variable is `null` or not

packages/inspection-report/src/components/Gallery.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const handleOnImageClick = useCallback((selectedImage) => {
```
souyahia-monk, line 123: rename this to `focusImage`

packages/inspection-report/src/components/Gallery.js
```diff
@@ -0,0 +1,180 @@
...
+        pictureList.map((image, index) => (
+          // eslint-disable-next-line react/no-array-index-key
+          <Thumbnail key={`${image.url}-${index}`} image={image} click={handleOnImageClick} />
+        ))
+      }
+
+      <Modal
+        animationType="slide"
+        transparent
+        visible={visibleImageInFullScreen}
```
souyahia-monk, line 142: for example, this could be `visible={!!focusedPhoto}`

packages/inspection-report/src/components/common/Thumbnail.js
```diff
@@ -0,0 +1,85 @@
...
+    color: '#fff',
+    fontSize: 14,
+    marginTop: 5,
+    textAlign: 'center',
+  },
+});
+
+function Thumbnail({ image, click }) {
+  const [imageSize, setImageSize] = useState(109);
+  const { width } = Dimensions.get('window');
```
souyahia-monk, line 23: replace with `const { width } = useWindowDimensions();`

packages/inspection-report/src/components/common/Thumbnail.js
```diff
@@ -0,0 +1,85 @@
...
+
+  useEffect(() => {
+    if (width < 480) {
+      setImageSize(109);
+    } else if (width > 480 && width < 768) {
+      setImageSize(140);
+    } else {
+      setImageSize(175);
+    }
+  }, [width]);
```
souyahia-monk, line 33: instead of doing this, you can remove the state and the useeffect and do something like : 

```javascript
// Declare this const OUTSIDE of the component, next to the styles
const responsiveImageSizes = [
  { breakpoint: 480, width: 109 },
  { breakpoint: 768, width: 140 },
  { breakpoint: null, width: 109 },
];

// Inside the component
const { width } = useWindowDimensions();
const imageSize = useMemo(
  () => responsiveImageSizes.find((r) => (!r.breakpoint || width <= r.breakpoint)).width,
  [width],
);
```

packages/inspection-report/src/components/common/Thumbnail.js
```diff
@@ -0,0 +1,85 @@
...
+            <Image
+              source={{
+                width: imageSize,
+                height: imageSize,
+                uri: image.url,
+              }}
+            />
+          )
+        }
+        <Text style={styles.text}>{image.label.en}</Text>
```
souyahia-monk, line 57: Don't forget to translate this

packages/inspection-report/src/components/Gallery.js
deepakglobant: We may sometimes have URLs with empty strings, which will throw us an error. so for that reason only I added an index here.

packages/inspection-report/src/components/common/Thumbnail.js
deepakglobant: Resolved all the comments.

packages/inspection-report/src/components/common/Thumbnail.js
```diff
@@ -0,0 +1,81 @@
...
+            <Image
+              source={{
+                width: imageSize,
+                height: imageSize,
+                uri: image.url,
+              }}
+            />
+          )
+        }
+        <Text style={styles.text}>{t(`gallery.parts.${image.label}`)}</Text>
```
souyahia-monk, line 59: No, if you take a look at the specifications in the Jira ticket, the translations are contained in the image given as a prop. The pictures look like this :

```typescript
interface ExamplePicture {
  height: number;
  width: number;
  url: string;
  mimetype: string;
  label?: {
    fr: string;
    en: string;
  };
}
```

So you should have something like :

```javascript
function Thumbnail({ image, click }) {
  const { i18n } = useTranslation();
  const label = useMemo(() => (image.label ? image.label[i18n.language] : ''), [i18n]);

  ...
    <Text>{label}</Text>
  ...
}
```

You can also remove all the translations of the car parts in the resources folder

deepakglobant: Done

souyahia-monk (approved):

PR #489, opened 2023-05-05
Feature/mn 255/overview component

souyahia-monk:


deepakglobant (approved):

yarn.lock
```diff
@@ -4091,16 +4091,11 @@
   resolved "https://registry.yarnpkg.com/@types/minimatch/-/minimatch-3.0.5.tgz#1001cc5e6a3704b83c236027e77f2f58ea010f40"
   integrity sha512-Klz949h02Gz2uZCMGwDUSDS1YBlTdDDgbWHi+81l29tQALUtvz4rAYi5uoVhE5Lagoq6DeqAUlbrHvW/mXDgdQ==
 
-"@types/node@*", "@types/node@~14.0.0":
+"@types/node@*", "@types/node@^17.0.29", "@types/node@~14.0.0":
```
rohitsharma120582, line 4094: Should we restrict to use the version of `@types/node` to be within the range of `~14.0.0` to avoid any issue instead of range from 14 to 17?

rohitsharma120582 (approved):

yarn.lock
souyahia-monk: I think it's fine since these version differences probably come from deep dependencies and not from our packages

PR #520, opened 2023-07-05
feat: create vehicle type selection (MN-297)

deepakglobant:
Added new workflow CAPTURE_VEHICLE_SELECTION for CAT application.
If vehicle type is not provided for CAT, then the user will be redirected to the vehicle type selection screen, else will redirect to the capture screen.

### DIFF
--- a/src/screens/CaptureVehicleSelection/index.js
+++ b/src/screens/CaptureVehicleSelection/index.js
@@ -1,0 +1,158 @@
+import { useMonitoring } from '@monkvision/corejs';
+import { useInterval } from '@monkvision/toolkit';
+import { Container } from '@monkvision/ui';
+import { useFocusEffect, useNavigation, useRoute } from '@react-navigation/native';
+import { LinearGradient } from 'expo-linear-gradient';
+import useSnackbar from 'hooks/useSnackbar';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { Text, View, useWindowDimensions } from 'react-native';
+import { Card, List, useTheme } from 'react-native-paper';
+import { useMediaQuery } from 'react-responsive';
+
+import { version } from '@package/json';
+import * as names from 'screens/names';
+import Artwork from '../Landing/Artwork';
+import LanguageSwitch from '../Landing/LanguageSwitch';
+import useGetInspection from '../Landing/useGetInspection';
+import { Clients, Workflows, useClient } from '../../contexts';
+import VehicleType from '../Landing/VehicleType';
+import useUpdateInspectionVehicle from '../Landing/useUpdateInspectionVehicle';
+import useZlibCompression from '../Landing/useZlibCompression';
+import styles from './styles';
+import { ClientParamMap } from '../paramsMap';
+
+export default function CaptureVehicleSelection() {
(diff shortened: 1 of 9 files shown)


src/screens/CaptureVehicleSelection/index.js
```diff
@@ -0,0 +1,164 @@
...
+import styles from './styles';
+
+const ClientParamMap = {
+  d9m: Clients.DEFAULT,
+  pd8: Clients.CAT,
+  '2bd': Clients.FASTBACK,
+  a3o: Clients.ALPHA,
+};
+
+export default function Landing() {
```
rohitsharma120582, line 31: Change the name of this function from `Landing` to `CaptureVehicalSelection`

src/screens/CaptureVehicleSelection/index.js
```diff
@@ -0,0 +1,164 @@
...
+import Artwork from '../Landing/Artwork';
+import LanguageSwitch from '../Landing/LanguageSwitch';
+import useGetInspection from '../Landing/useGetInspection';
+import { Clients, Workflows, useClient } from '../../contexts';
+import VehicleType from '../Landing/VehicleType';
+import useUpdateInspectionVehicle from '../Landing/useUpdateInspectionVehicle';
+import useZlibCompression from '../Landing/useZlibCompression';
+import styles from './styles';
+
+const ClientParamMap = {
```
rohitsharma120582, line 24: The `screens/landing` file also has `ClientParamMap` object. Please create a new file `screens/paramsMap.js` for having `ClientParamMap` and `VehicleTypeMap` definitions. This will serve as a common file to export these objects and use them in both `landing` page and this page.

src/screens/Landing/index.js
```diff
@@ -105,6 +105,19 @@ export default function Landing() {
           vehicle: { vehicleType: VehicleTypeMap[vehicleTypeParam] ?? 'cuv' },
         },
       );
+    } else if (workflow === Workflows.CAPTURE_VEHICLE_SELECTION && !route.params?.captureComplete) {
+      if (vehicleTypeParam) {
```
rohitsharma120582, line 109: Show `CAPTURE_VEHICLE_SELECTION` screen in following scenarios
* If `v` param is not in the path and `vehicleTypeParam` value is undefined
* If `v` param is in the path, but `vehicleTypeParam` value is invalid

src/screens/Landing/index.js
```diff
@@ -84,7 +84,7 @@ export default function Landing() {
 
     return {
       inspectionId: urlParams.get('i'),
-      vehicleTypeParam: urlParams.get('v') ?? 1,
+      vehicleTypeParam: urlParams.get('v'),
```
rohitsharma120582, line 87: Change it with `vehicleTypeParam: (clientParam === 'pd8') ? urlParams.get('v') : urlParams.get('v') ?? 1`

If client is not `CAT` and `v` param is not in path, then default vehicle should be `cuv`.

rohitsharma120582 (changes requested):

src/screens/Landing/index.js
deepakglobant: Done.

rohitsharma120582 (approved):

PR #535, opened 2023-07-28
fix: add dependency of config in monitoring provider

deepakglobant:


rohitsharma120582 (approved):

rohitsharma120582 (approved):

rohitsharma120582:
Need to create new projects in new Sentry dashboard and use their DSN name in the app.

rohitsharma120582 (approved):

PR #536, opened 2023-07-28
feat: add generic monitoring (MN-299)

deepakglobant:


app.config.js
```diff
@@ -280,6 +280,6 @@ export default {
       palette,
     },
     ...app,
-    ...eas.build.preview.env,
+    ...eas.build.development.env,
```
rohitsharma120582, line 283: Point to `preview` env as this build will be deployed on Preview environment

src/main.js
```diff
@@ -1,32 +1,30 @@
-import React from 'react';
+import React, { useEffect } from 'react';
```
rohitsharma120582, line 1: `useEffect` is not being used at the component level. Plz remove it.

rohitsharma120582 (changes requested):

src/main.js
deepakglobant: Resolved.

app.config.js
deepakglobant: Resolved.

rohitsharma120582 (approved):

souyahia-monk (approved):

PR #549, opened 2023-08-11
Created the common package

souyahia-monk:


packages/common/.eslintignore
```diff
@@ -0,0 +1,10 @@
+dist
```
deepakglobant, line 1: we might only need `lib` folder here.

packages/common/.gitignore
```diff
@@ -0,0 +1,7 @@
+dist
```
deepakglobant, line 1: also we will not need dist here

packages/monitoring/src/react/hooks.ts
```diff
@@ -5,6 +5,6 @@ import { MonitoringContext } from './context';
 /**
  * Custom hook that allows you to access the Monitoring Context methods inside a component.
  */
-export function hooks(): MonitoringAdapter {
+export function useMonitoring(): MonitoringAdapter {
```
deepakglobant, line 8: I have made these changes in my pr and it's now merged. So we can remove this one.

deepakglobant (changes requested):

deepakglobant (approved):

PR #554, opened 2023-08-22
feat: (MN-338) allow multiple currencies

deepakglobant:


packages/inspection-report/src/components/DamageReport/DamageManipulator.js
```diff
@@ -269,6 +271,7 @@ DamageManipulator.propTypes = {
 DamageManipulator.defaultProps = {
   damage: undefined,
   damageMode: DamageMode.ALL,
+  currencyCharacter: '',
```
souyahia-monk, line 274: switch default to "$"

packages/inspection-report/src/components/DamageReport/Overview/index.js
```diff
@@ -182,6 +185,7 @@ Overview.propTypes = {
 Overview.defaultProps = {
   damageMode: DamageMode.ALL,
   damages: [],
+  currencyCharacter: '',
```
souyahia-monk, line 188: make the prop required

packages/inspection-report/src/components/DamageReport/UpdateDamagePopUp/index.js
```diff
@@ -236,6 +239,7 @@ UpdateDamagePopUp.propTypes = {
 UpdateDamagePopUp.defaultProps = {
   damage: undefined,
   damageMode: DamageMode.ALL,
+  currencyCharacter: '',
```
souyahia-monk, line 242: make the prop required

src/screens/InspectionCapture/index.js
```diff
@@ -16,6 +16,7 @@ import useSnackbar from '../../hooks/useSnackbar';
 import useFullscreen from './useFullscreen';
 
 const enableComplianceCheck = false;
+const currencyCharacter = '€';
```
souyahia-monk, line 19: instead of doing this and passing the character in the route, you can hardcode the prop that is given to the inspectionreport component

packages/inspection-report/src/components/DamageReport/DamageReport.js
```diff
@@ -101,6 +101,7 @@ export default function DamageReport({
   inspectionId,
   vehicleType,
   damageMode,
+  currencyCharacter,
```
souyahia-monk, line 104: didnt you forgot to add the prop to the proptypes? If so, make it optional with the "$" as a default value

deepakglobant: Resolved all the comments

rohitsharma120582 (commented):
What do you think @deepakglobant and @souyahia-monk to create and use a context at `DamageReport` component level to set the currency value based on the received prop and use it at different levels to avoid multi level props drilling.

souyahia-monk:
> What do you think @deepakglobant and @souyahia-monk to create and use a context at `DamageReport` component level to set the currency value based on the received prop and use it at different levels to avoid multi level props drilling.

Good idea!

packages/inspection-report/src/hooks/useCurrency.js
```diff
@@ -0,0 +1,44 @@
...
+export function CurrencyProvider({ children }) {
+  const [currency, setCurrency] = useState('$');
+
+  /**
+   * Set current currency.
+   *
+   * @param value Value of currency
+   * @return {void}
+  */
+  const updateCurrency = useCallback((value) => setCurrency(value), []);
```
rohitsharma120582, line 25: No need to to create this function and use `useCallback` here. You can simple add `updateCurrency: setCurrency` in context value.

rohitsharma120582 (changes requested):

packages/inspection-report/src/hooks/useCurrency.js
deepakglobant: Done

rohitsharma120582 (approved):

souyahia-monk (approved):

PR #610, opened 2023-11-21
Refacto/mn 408/queue system

souyahia-monk:


packages/common/README/UTILITIES.md
```diff
@@ -31,9 +60,19 @@ import { permutations } from '@monkvision/common';
...
-| Param     | Type                                                   | Description                                         | Required | Default Value |
-|-----------|--------------------------------------------------------|-----------------------------------------------------|----------|---------------|
-| array     | array                                                  | The array to compute the permutations of.           | ✔️       |               |
+# Queue Utils
+## useQueue
+```typescript
+import { useQueue } from '@monkvision/common';
+
+function TestComponent() {
+  const queue = useQueue<string>((item) => console.log(item));
```
rohitsharma120582, line 71: Make a change in this statement for "process" function as function should return a Promise

PR #648, opened 2023-12-13
MN-430: PhotoHUDPreview with SightsSlider + Counter

dlymonk:


packages/public/common/src/hooks/useObjectTranslation.ts
```diff
@@ -0,0 +1,24 @@
+import { LabelTranslation } from '@monkvision/types';
+import { useTranslation } from 'react-i18next';
+
+export function useObjectTranslation(): {
+  tObj: (label: LabelTranslation) => string;
+} {
```
souyahia-monk, line 6: Define a type for this return value like maybe `UseObjectTranslationResult`

packages/public/common/src/hooks/useObjectTranslation.ts
```diff
@@ -0,0 +1,24 @@
+import { LabelTranslation } from '@monkvision/types';
+import { useTranslation } from 'react-i18next';
+
+export function useObjectTranslation(): {
```
souyahia-monk, line 4: Add documentation for this hook

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUD.tsx
```diff
@@ -0,0 +1,13 @@
+import { labels } from '@monkvision/sights';
+import { PhotoCaptureHUDPreview } from './PhotoCaptureHUDPreview';
+import { PhotoCaptureHUDButtons } from './PhotoCaptureHUDButtons';
+import { styles } from './PhotoCaptureHUD.styles';
+
+export function PhotoCaptureHUD() {
```
souyahia-monk, line 6: Add TSDOC for this component

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDButtons/PhotoCaptureHUDButtons.tsx
```diff
@@ -3,7 +3,7 @@ import { Icon, TakePictureButton } from '@monkvision/common-ui-web';
 import { useInteractiveStatus } from '@monkvision/common';
 import { useCaptureHUDButtonsStyles } from './hooks';
 
-export interface CaptureHUDButtonsProps {
+export interface PhotoCaptureHUDButtonsProps {
```
souyahia-monk, line 6: Add TSDoc for this interface

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreview/PhotoCaptureHUDPreview.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { PhotoCaptureHUDSightsSlider } from './PhotoCaptureHUDSightsSlider';
+import { PhotoCaptureHUDCounter } from './PhotoCaptureHUDSightsCounter';
+
+export interface PhotoCaptureHUDPreviewProps {
+  sights: LabelDictionary;
+  onAddDamage?: () => void;
+}
+
+export function PhotoCaptureHUDPreview({ sights, onAddDamage }: PhotoCaptureHUDPreviewProps) {
+  const [currentSight, setCurrentSight] = useState('');
```
souyahia-monk, line 16: currentSight should be taken as a prop by this component

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreview/PhotoCaptureHUDPreview.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useObjectTranslation } from '@monkvision/common';
+import { useTranslation } from 'react-i18next';
+import { LabelDictionary } from '@monkvision/types';
+import { Button } from '@monkvision/common-ui-web';
+import { PhotoCaptureHUDPreview, usePhotoCaptureHUDPreview } from './hook';
+import { PhotoCaptureHUDSightsSlider } from './PhotoCaptureHUDSightsSlider';
+import { PhotoCaptureHUDCounter } from './PhotoCaptureHUDSightsCounter';
+
+export interface PhotoCaptureHUDPreviewProps {
+  sights: LabelDictionary;
```
souyahia-monk, line 11: The `sights` prop should actually be a list of sights to display at the bottom. The type of this prop should be `Sight[]`

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreview/PhotoCaptureHUDPreview.tsx
```diff
@@ -0,0 +1,42 @@
...
+  onAddDamage?: () => void;
```
souyahia-monk, line 12: You should also have the following props :

`onSelectSight?: () => void;`

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreview/PhotoCaptureHUDPreview.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  useEffect(() => {
+    // Set the first sight as the default value for the current sight
+    setSightsArray(Object.values(sights).map((label) => tObj(label)));
+  }, [i18n.language]);
+
+  return (
+    <div style={style.containerStyle}>
+      <div style={style.top}>
+        <PhotoCaptureHUDCounter sight={sightsArray} styles={style} />
```
souyahia-monk, line 31: Don't pass the styles as a prop to this component, declare each component's style next to it in a file

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreview/PhotoCaptureHUDPreview.tsx
```diff
@@ -0,0 +1,42 @@
...
+        <Button onClick={onAddDamage}>+ Damage</Button>
```
souyahia-monk, line 32: Instead of passing a "+" character, use the existing icon like that :

```tsx
<Button icon='add' onClick={onAddDamage}>Damage</Button>
```

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreview/PhotoCaptureHUDPreview.tsx
```diff
@@ -0,0 +1,42 @@
...
+      </div>
+      <PhotoCaptureHUDSightsSlider
+        sight={sightsArray}
+        currentSight={currentSight}
+        onSightSelected={(sight) => setCurrentSight(sight)}
+        styles={style}
```
souyahia-monk, line 38: Same as before, each component has its own styles

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreview/PhotoCaptureHUDSightsCounter.tsx
```diff
@@ -0,0 +1,19 @@
+import { PhotoCaptureHUDPreview } from './hook';
+
+export interface PhotoCaptureHUDCounterProps {
+  sightDisable?: string[];
```
souyahia-monk, line 4: rename this prop to `sightsTaken`

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreview/PhotoCaptureHUDSightsCounter.tsx
```diff
@@ -0,0 +1,19 @@
...
+  sight: string[];
```
souyahia-monk, line 5: change this prop to `totalSights: number`

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreview/PhotoCaptureHUDSightsSlider.tsx
```diff
@@ -0,0 +1,35 @@
+import { Button } from '@monkvision/common-ui-web';
+import { PhotoCaptureHUDPreview } from './hook';
+
+export interface SliderProps {
+  sight: string[];
```
souyahia-monk, line 5: change this prop to `sights: Sight[]`

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreview/PhotoCaptureHUDSightsSlider.tsx
```diff
@@ -0,0 +1,35 @@
...
+  currentSight: string;
+  onSightSelected: (sight: string) => void;
```
souyahia-monk, line 7: change signature to `(sight: Sight) => void`

packages/public/common/src/hooks/useObjectTranslation.ts
```diff
@@ -0,0 +1,26 @@
...
+import { useTranslation } from 'react-i18next';
+
+/**
+ * The result of the useObjectTranslation. It contains a function which takes a LabelTranslation object and return the
+ * translated label sync with the actual selected language.
+ */
+export interface UseObjectTranslationResult {
+  /**
+   * Function translating a LabelTranslation object into a translated label sync with the actual selected language.
+   * @param obj
```
souyahia-monk, line 11: Remove this @param annotation

packages/public/common/src/hooks/useObjectTranslation.ts
```diff
@@ -0,0 +1,26 @@
...
+   */
+  tObj: (obj: TranslationObject) => string;
+}
+
+/**
+ * Custom hook used to get the label with the actual selected language.
```
souyahia-monk, line 17: Change doc to : `Custom hook used to get a translate function (\`tObj\`) that translates TranslationObjects.`

packages/public/common/src/hooks/useSightLabel.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * The result of the useSightLabel. It contains a function which takes a Sight object and return the
+ * translated label sync with the actual selected language.
+ */
+export interface UseSightLabelResult {
+  label: (sight: Sight) => string;
+}
+
+export interface UseSightLabelParams {
```
souyahia-monk, line 12: Add TSDoc for this interface

packages/public/common/src/hooks/useSightLabel.ts
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Function translating a Sight object into a translated label sync with the actual selected language.
+   * @param obj
```
souyahia-monk, line 15: Remove this @param annotation

souyahia-monk, line 15: Update the TSDoc for this field (it is not a function), and remove the @param annotation

packages/public/common/src/hooks/useSightLabel.ts
```diff
@@ -0,0 +1,31 @@
...
+   */
+  labels: LabelDictionary;
+}
+
+/**
+ * Custom hook used to get the label with the actual selected language.
```
souyahia-monk, line 21: ...to get the label ***of a sight*** with...

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUD.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const onScrollToSelected = (index: number, smooth: boolean): void => {
+    if (ref.current && ref.current.children.length > index) {
+      ref.current.children[index].scrollIntoView({
+        behavior: smooth ? 'smooth' : ('instant' as ScrollBehavior),
+        inline: 'center',
+      });
+      setCurrentSightSliderIndex(index);
+    }
+  };
```
souyahia-monk, line 35: This "scrollToSelected" logic should be moved into the Slider component. It is the job of the Slider component to decide how and when to scroll its content.

dlymonk: It was initially located down in the slider component, but I moved it up to the AddDamage component to enable scrolling to the last selected component when returning to the preview. Or should I create a other function which will be used in AddDamage return button?

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreviewSight/PhotoCaptureHUDPreview.styles.ts
```diff
@@ -0,0 +1,62 @@
+import { Styles } from '@monkvision/types';
+
+export const styles: Styles = {
```
souyahia-monk, line 3: Change the file structure from this :
```
PhotoCaptureHUDPreviewSight
├── PhotoCaptureHUDPreview.styles.ts
├── PhotoCaptureHUDPreview.tsx
└── components
    ├── PhotoCaptureHUDSightsCounter.tsx
    ├── PhotoCaptureHUDSightsOverlay.tsx
    └── PhotoCaptureHUDSightsSlider.tsx
```
to this :
```
PhotoCaptureHUDPreview
├── PhotoCaptureHUDPreview.styles.ts
├── PhotoCaptureHUDPreview.tsx
├── PhotoCaptureHUDSightCounter
│   ├── index.ts
│   ├── PhotoCaptureHUDSightCounter.styles.ts
│   └── PhotoCaptureHUDSightCounter.tsx
├── PhotoCaptureHUDSightOverlay
│   ├── index.ts
│   ├── PhotoCaptureHUDSightOverlay.styles.ts
│   └── PhotoCaptureHUDSightOverlay.tsx
└── PhotoCaptureHUDSightSlider
    ├── index.ts
    ├── PhotoCaptureHUDSightSlider.styles.ts
    └── PhotoCaptureHUDSightSlider.tsx
```

Each component should have its own style declarations! I would rather have 30 small style files than 1 huuuuuge style file

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreviewSight/PhotoCaptureHUDPreview.tsx
```diff
@@ -0,0 +1,58 @@
...
+export interface PhotoCaptureHUDPreviewProps {
+  sights?: Sight[];
+  currentSight?: Sight;
+  onSightSelected?: (sight: Sight) => void;
+  onAddDamage?: (state: boolean) => void;
+  sightsTaken?: Sight[];
+  onScrollToSelected?: (index: number, smooth: boolean) => void;
+}
+
+export const PhotoCaptureHUDPreview = forwardRef<HTMLDivElement, PhotoCaptureHUDPreviewProps>(
```
souyahia-monk, line 18: No need to forwardRef here

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreviewSight/PhotoCaptureHUDPreview.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const style = usePhotoCaptureHUDPreview();
+
+    return (
+      <div style={style.containerStyle}>
+        <PhotoCaptureHUDSightsOverlay sight={currentSight} />
+        <div style={style.top}>
+          <PhotoCaptureHUDCounter totalSights={sights?.length} sightsTaken={sightsTaken?.length} />
+          <Button
+            icon='add'
+            onClick={() => onAddDamage?.(true)}
```
souyahia-monk, line 39: remove the dot here

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreviewSight/PhotoCaptureHUDPreview.tsx
```diff
@@ -0,0 +1,58 @@
...
+            primaryColor='secondary-xdark'
+            data-testid='monk-test-btn'
+            style={style.addDamageButton}
+          >
+            Damage
```
souyahia-monk, line 44: Translate this!

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreviewSight/components/PhotoCaptureHUDSightsCounter.tsx
```diff
@@ -0,0 +1,19 @@
+import { usePhotoCaptureHUDPreview } from '../hook';
+
+export interface PhotoCaptureHUDCounterProps {
+  totalSights?: number;
+  sightsTaken?: number;
```
souyahia-monk, line 5: you can make these fields required and remove the 'error-no-sight' label

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreviewSight/components/PhotoCaptureHUDSightsOverlay.tsx
```diff
@@ -0,0 +1,12 @@
+import { Sight } from '@monkvision/types';
+import { SightOverlay } from '@monkvision/common-ui-web';
+import { usePhotoCaptureHUDPreview } from '../hook';
+
+export interface PhotoCaptureHUDSightsOverlayProps {
+  sight?: Sight;
```
souyahia-monk, line 6: You can make this field required and remove the ternary operator below

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreviewSight/components/PhotoCaptureHUDSightsSlider.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export interface PhotoCaptureHUDSightsSliderProps {
+  sights?: Sight[];
+  currentSight?: string;
+  sightsTaken?: Sight[];
+  onSightSelected?: (sight: Sight) => void;
+  onScrollToSelected?: (index: number, smooth: boolean) => void;
+}
+
+export const PhotoCaptureHUDSightsSlider = forwardRef<
```
souyahia-monk, line 16: No need to forwardRef here

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreviewSight/components/PhotoCaptureHUDSightsSlider.tsx
```diff
@@ -0,0 +1,62 @@
...
+              onClick={() => {
+                onSightSelected(sight);
+                onScrollToSelected(index, true);
+              }}
+              data-testid={`sight-btn-${index}`}
+            >
+              {label(sight)}
+            </Button>
+          );
+        })}
```
souyahia-monk, line 58: move this outside of the JSX! We try to limit the quantity of typescript code inside JSX. You can create a component called "SliderButton" inside the same file (this component must NOT be exported) and simply map your sights to this component

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDPreviewSight/components/PhotoCaptureHUDSightsSlider.tsx
```diff
@@ -0,0 +1,62 @@
...
+    ref,
+  ) => {
+    const { label } = useSightLabel({ labels });
+    const style = usePhotoCaptureHUDPreview();
+
+    return (
+      <div style={style.slider} ref={ref}>
+        {sights?.map((sight, index) => {
+          let primaryColor = 'secondary-xdark';
+          let icon = '' as IconName;
```
souyahia-monk, line 37: `let icon = undefined as IconName | undefined;`

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUD.tsx
souyahia-monk: For me, your Slider component should have a useEffect that is called every time the currentSight changes that automatically slides the content of the slider. If you have this, normally when you come back from the Add Damage view, the slider component will be placed again in the DOM tree and the useEffect will be called automatically

packages/public/inspection-capture-web/src/PhotoCapture.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
```
souyahia-monk, line 1: You can remove this import

packages/public/inspection-capture-web/src/PhotoCapture.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Camera, CameraHUDProps, MonkPicture } from '@monkvision/camera-web';
+import { Sight } from '@monkvision/types';
+import { PhotoCaptureHUD } from './PhotoCaptureHUD';
+import { styles } from './PhotoCapture.styles';
+import { useCameraConfig } from './hooks/useCameraConfig';
+
+export interface PhotoCaptureProps {
```
souyahia-monk, line 8: Add TSDoc for this interface

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUD.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { PhotoCaptureHUDSightPreview } from './PhotoCaptureHUDSightPreview';
+import { HUDMode, usePhotoCaptureHUD } from './hook';
+import { useSightState, useHUDMode } from '../hooks';
+import { i18nAddDamage } from './i18n';
+
+export interface PhotoCaptureHUDProps extends CameraHUDProps {
+  sights: Sight[];
+}
+
+export const PhotoCaptureHUD = i18nWrap(
```
souyahia-monk, line 16: put the i18nWrap in the PhotoCapture component instead

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDAddDamagePreview/PhotoCaptureHUDAddDamagePreview.styles.ts
```diff
@@ -0,0 +1,26 @@
...
+    alignSelf: 'flex-start',
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+    margin: '10px',
+    zIndex: '1',
+    top: '0',
+    right: '0',
+  },
+  addDamageButton: {
+    backgroundColor: 'rgba(52, 53, 63, 0.64)',
```
souyahia-monk, line 24: Use a color from the theme instead (you don't need to have the EXACT same color as the one in the Figma, juste take whatever color is the closest in the default palette). If you want to change the opacity of the color you can use the utility function that does this in the common package

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDSightPreview/AddDamageButton/AddDamageButton.styles.ts
```diff
@@ -0,0 +1,7 @@
+import { Styles } from '@monkvision/types';
+
+export const styles: Styles = {
+  button: {
+    backgroundColor: 'rgba(52, 53, 63, 0.64)',
```
souyahia-monk, line 5: Same as before, use a color from the theme

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDSightPreview/SightsCounter/SightsCounter.styles.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const styles: Styles = {
+  counter: {
+    display: 'flex',
+    color: 'white',
+    alignItems: 'center',
+    justifyContent: 'center',
+    padding: '6px 12px',
+    borderRadius: '8px',
+    backgroundColor: 'rgba(28, 28, 30, 0.64)',
```
souyahia-monk, line 11: Same as before : use color from theme

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDSightPreview/SightsSlider/SightsSlider.tsx
```diff
@@ -0,0 +1,98 @@
...
+    icon = 'check';
+  }
+  return (
+    <Button
+      style={styles['button']}
+      icon={icon}
+      primaryColor={primaryColor}
+      onClick={() => {
+        onSightSelected(sight);
+      }}
```
souyahia-monk, line 50: Remove brackets : `onClick={() => onSightSelected(sight)}`

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDSightPreview/SightsSlider/SightsSlider.tsx
```diff
@@ -0,0 +1,98 @@
...
+      data-testid={`sight-btn-${sight.id}`}
+    >
+      {label(sight)}
+    </Button>
+  );
+}
+
+const onScrollToSelected = (
```
souyahia-monk, line 58: From what I understand, this function **actually** scrolls to the selected sight, so you can rename it to `scrollToSelectedChild` or something like that

packages/public/inspection-capture-web/src/PhotoCaptureHUD/hook.ts
```diff
@@ -0,0 +1,35 @@
...
+   */
+  ADD_DAMAGE = 'add-damage',
+}
+
+export interface PhotoCaptureHUD {
+  container: CSSProperties;
+  previewContainer: CSSProperties;
+}
+
+export function usePhotoCaptureHUD(): PhotoCaptureHUD {
```
souyahia-monk, line 24: Rename this hook to `usePhotoCaptureHUDStyle`

packages/public/inspection-capture-web/src/PhotoCaptureHUD/hook.ts
```diff
@@ -0,0 +1,35 @@
...
+   * Default mode
+   */
+  DEFAULT = 'default',
+  /**
+   * Add damage mode/preview
+   */
+  ADD_DAMAGE = 'add-damage',
+}
+
+export interface PhotoCaptureHUD {
```
souyahia-monk, line 19: Rename this interface to `PhotoCaptureHUDStyle`

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDSightPreview/hook.ts
```diff
@@ -0,0 +1,25 @@
+import { CSSProperties } from 'react';
+import { styles } from './PhotoCaptureHUDSightPreview.styles';
+import { useCameraConfig } from '../../hooks/useCameraConfig';
+
+export interface PhotoCaptureHUDSightPreview {
```
souyahia-monk, line 5: Rename this interface to `PhotoCaptureHUDSightPreviewStyle`

packages/public/inspection-capture-web/src/PhotoCaptureHUD/translations/de.json
```diff
@@ -0,0 +1,4 @@
+{
```
souyahia-monk, line 1: Put all translations in the `src/translations` directory instead of in the component directory itself : we should have only one translation file per language in the whole package.

For the format of the translation keys, there are two schools :
- Devs who prefer to have one translation key per word or phrase (ex "ok" or "message.hello") and re-use the same translation keys everytime the same message is used in the app
- Devs who prefer to have one translation key per translated text element on the screen, meaning that if you have 2 buttons saying "ok" you will have 2 different translation keys with the same value

In my case, I prefer the second option. Even though it means having a lot of different keys for the same words sometime, I think it's easier to maintain. Plus, if tomorrow the product team asks us to change the content of the button from "OK" to "YES", we just have to change the translation file instead of the react file.

Anyway, this means that in order to properly implement the second option, and if we want to have all the translations in on single file, the keys should be named something like that : `baseComponent.childComponent.otherChildComponent.element`

Of course, we don't have to put the entire names of each and every component, we can be a bit flexible and prefer readability. In your case, you could do something like :

```json
{
  "photo": {
    "hud": {
      "sight": {
        "addDamageBtn": "Schaden"
      },
      "addDamage": {
        "cancelBtn": "stornieren"
      }
    }
  }
}
`
``

I hope all of this is clear, if you have any question you can ask me on Slack

File: packages/public/inspection-capture-web/src/hooks/useCameraConfig.ts [hunk 42]
```diff
@@ -0,0 +1,27 @@
+import { useState } from 'react';
+import { CameraFacingMode, CameraResolution, CompressionFormat } from '@monkvision/camera-web';
+
+export function useCameraConfig() {
```
souyahia-monk, line 4: Okay so I think you created this hook to mimic how I implemented it in the TestView component in the test app. However I did this just to test the component. In reality we don't need all of this.

I think you can remove this hook and have all the camera config be passed as props to the PhotoCapture component (for now, maybe we will change this later)

packages/public/inspection-capture-web/src/hooks/useHUDMode.ts
```diff
@@ -0,0 +1,11 @@
+import { useState } from 'react';
+import { HUDMode } from '../PhotoCaptureHUD/hook';
+
+export function useHUDMode() {
```
souyahia-monk, line 4: I think this hook is useless. You can directly do this in the component :

```tsx
export function PhotoCaptureHUD(...) {
  const [mode, setMode] = useState<HUDMode>(HUDMode.DEFAULT);
  ...
    <PhotoCaptureHUDAddDamagePreview onCancel={() => setMode(HUDMode.DEFAULT)} />
  ...
    <PhotoCaptureHUDSightPreview onAddDamage={() => setMode(HUDMode.ADD_DAMAGE)} />
  ...
}
```

packages/public/inspection-capture-web/src/hooks/useSightState.ts
```diff
@@ -0,0 +1,40 @@
+import { Sight } from '@monkvision/types';
+import { useState } from 'react';
+
+/**
+ * Custom hook used to initialize and manipulate sight state.
+ */
+export function useSightState(sights: Sight[]) {
+  const [sightSelected, setSightSelected] = useState<Sight>(sights[0]);
```
souyahia-monk, line 8: rename this to `[selectedSight, setSelectedSight]`

packages/public/inspection-capture-web/src/hooks/useSightState.ts
```diff
@@ -0,0 +1,40 @@
...
+      const sightsNotTaken = sights.filter((sight) => !updatedSightsTaken.includes(sight));
+      const nextSight = sightsNotTaken[0];
+
+      if (nextSight) {
+        handleSightSelected(nextSight);
+      }
+
+      return updatedSightsTaken;
+    });
+  };
```
souyahia-monk, line 32: Hmm I don't really like calling `setState` inside another `setState` callback. It gets messy and it's hard to read. I think. I think you could change this function to :

```typescript
const handleSightTaken = () => {
  setSightsTaken((value) => [...value, selectedSight]);
  const nextSight = sights.filter((sight) => !updatedSightsTaken.includes(sight))[0];
  if (nextSight) {
    setSelectedSight(nextSight);
  }
};
```

Also one thing to note : you don't have to specify the return type when declaring arrow functions, I think it is overkill

packages/public/inspection-capture-web/test/PhotoCaptureHUD/PhotoCaptureHUDSightPreview/AddDamageButton.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { Button } from '@monkvision/common-ui-web';
+import { useTranslation } from 'react-i18next';
+import { expectPropsOnChildMock } from '@monkvision/test-utils';
+import { AddDamageButton } from '../../../src/PhotoCaptureHUD/PhotoCaptureHUDSightPreview/AddDamageButton';
+
+describe('AddDamage component', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+  it('should get passed onAddDamage callback', () => {
```
souyahia-monk, line 20: add new line before this line

packages/public/inspection-capture-web/test/hooks/useHUDMode.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+    unmount();
+  });
+
+  it('should return a empty array of sights as SightTaken default state', () => {
+    const { result, unmount } = renderHook(() => useHUDMode());
+
+    expect(typeof result.current.handleAddDamage).toEqual('function');
+    unmount();
+  });
+  it('should update mode with handleAddDamage', () => {
```
souyahia-monk, line 20: add new line before this line

packages/public/inspection-capture-web/test/hooks/useSightState.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  it('should initialize with the first sight as selected and an empty list of taken sights', () => {
+    const { result, unmount } = renderHook(() => useSightState(sights));
+
+    expect(result.current.sightSelected).toEqual(sights[0]);
+    expect(result.current.sightsTaken).toEqual([]);
+    expect(typeof result.current.handleSightTaken).toEqual('function');
+    expect(typeof result.current.handleSightSelected).toEqual('function');
+    unmount();
+  });
+  it('should update sightSelected when handleSightSelected is called', () => {
```
souyahia-monk, line 26: add new line before this line

packages/public/inspection-capture-web/src/PhotoCaptureHUD/translations/de.json
dlymonk: Super clear I really appreciate your deep explanation ! Thanks

packages/public/inspection-capture-web/src/hooks/useCameraConfig.ts
dlymonk: I created this hook to also get the CameraDimensions which is used in the hook SightPreviewStyle. 
The goal is to assign the right aspectRatio for the SightOverlay. (Instead of '16/9' that I have implemented previously)

Should we keep this useCameraConfig for that?

packages/public/inspection-capture-web/src/hooks/useSightState.ts
dlymonk: yeah I agree, lack of readability! 

So I found an other way to update `SelectedSight` state after being sure that `setSightsTaken` have really updated his state:
```
useEffect(() => {
    const nextSight = sights.filter((sight) => !sightsTaken.includes(sight))[0];
    if (nextSight) {
      setSelectedSight(nextSight);
    }
  }, [sightsTaken]);
```
Don't you think that `useSelectedSight` hook is now a bit useless ? Should I move the code to `PhotoCaptureHUD` which is the only place that we use it?

packages/public/inspection-capture-web/src/PhotoCaptureHUD/PhotoCaptureHUDAddDamagePreview/PhotoCaptureHUDAddDamagePreview.styles.ts
dlymonk: I added it directly to the component like this:
```
export function AddDamageButton({ onAddDamage }: AddDamageButtonProps) {
  const { t } = useTranslation();
  const { palette } = useMonkTheme();

  const bgColor = getHexFromRGBA({ ...getRGBAFromString(palette.secondary.xdark), a: 0.64 });

 return (
    <Button
      icon='add'
      onClick={() => onAddDamage(HUDMode.ADD_DAMAGE)}
      data-testid='monk-test-btn'
      style={{ backgroundColor: bgColor }}
    >
      {t('photo.hud.sight.addDamageBtn')}
    </Button>
  );
}
```
Should I create a hook file to handle this for every component ?

souyahia-monk: Okay so three things :
- First, know that the two util functions `getHexFromRGBA` and `getRGBAFromString` are a tiny bit expensive so whenever you use them (or use another function that uses them), I think it's better to memoize them (put them in `useMemo`). So in your case it would mean something like that :
```typescript
const bgColor = useMemo(
  () => getHexFromRGBA({ ...getRGBAFromString(palette.secondary.xdark), a: 0.64 }),
  [palette],
);
```
- Second : in [this PR](https://github.com/monkvision/monkjs/pull/656), I created a util function called `changeAlpha` that changes the alpha value of a color, which is exactly what you did here. Since I already wrote the function, the doc and the tests in the PR, I suggest we merge the PR, and then you simply rebase the changes onto your branch and use the util function instead of manually changing the alpha value.
- Third : For your custom hook suggestion, if you are using the exact same alpha change for every semi-transparent button in the HUD, you could definitely create a custom hook called `usePhotoHUDButtonBackground` that will do two things :
  - Change the alpha of the color from the palette and return the proper background color
  - Make sure to memoize the call to `changeAlpha` since its an expensive function

What do you think?

packages/public/inspection-capture-web/src/hooks/useSightState.ts
souyahia-monk: Hmm I don't understand why the first solution I suggested isn't good enough? Having a random useEffect in the component that could simply be replaced by a few lines of codes in the handler function isn't really optimal so could you explain to me why you need this?

packages/public/inspection-capture-web/src/hooks/useCameraConfig.ts
souyahia-monk: Oooooh okay I misread the hook, I think your approach is wrong, I'll explain it to you by voice later it will be easier

souyahia-monk: Ooooh okay so I misread the hook my bad, I think your approach is wrong. Let me explain to you how the Camera works internally so that you'll get a better grasp at how things are working :

- We pass some `CameraConfig` properties as props to the `Camera` component. These properties allow us to configure some properties of the camera such as the resolution, the orientation (front / rear camera) etc.
- When the `Camera` component receives these `CameraConfig` values, it converts them into `MediaConstraints` using the `getMediaConstraints` function (I will explain to you what are `MediaConstraints` later).
- Then, the `Camera` component will take these `MediaConstraints` and pass them to the `useUserMedia` hook. This hook is simply a wrapper hook, that wraps the `getUserMedia` function and adds React logic like state management etc to it.
- The `getUserMedia` function is a Vanilla JavaScript function available on all browsers ([here is the official doc](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia)). This function does the following : 
  - It accepts `MediaConstraints` as parameters. These constraints are simply Media requirements that we ask to the browser (for instance, we ask the browser "_hey, can you give me access to the front camera, with a 1080p resolution please?_")
  - It then asks the browser to give us the `MediaStream` object (a media stream for the device, like a camera stream or audio stream) and it returns the `MediaStream`
- The `Camera` component then binds the resulting `MediaStream` (which corresponds to the device camera stream) to an HTML `video` element to display the camera stream onto the screen.

Now that you understand how the `Camera` component works, note one things : the `MediaConstraints` are only "requirements" that we ask to the browser, but they are not necessarily the EXACT properties of the resulting stream that we will get. For instance, what happens if we ask the browser for a 4K UHD camera stream when the device only has a 1080p camera? Or when we ask for a rear camera when the device is a laptop and it only has a front facing camera? Well, the answer is : it depends on the browser 😂 Sometimes, the browser will crop and scale the original camera feed to make it match the requirements, sometimes it will completely ignore our requirements and directly give us the original camera feed. Basically the take away from all of this is : the specs we pass to the `Camera` component are only "requirements", but not necessarily the resulting properties.

packages/public/inspection-capture-web/src/hooks/useSightState.ts
souyahia-monk:
```typescript
const handleSightTaken = () => {
  const updatedSightsTaken = [...sightsTaken, selectedSight];
  setSightsTaken(updatedSightsTaken);
  const nextSight = sights.filter((sight) => !updatedSightsTaken.includes(sight))[0];
  if (nextSight) {
    setSelectedSight(nextSight);
  }
};
```

packages/public/inspection-capture-web/src/hooks/useCameraConfig.ts
souyahia-monk: The `Camera` component should pass the stream dimensions in the camera handle to the HUD :

- Dans `CameraHUD.types.ts` : rajouter un champ `streamDimensions` à l'interface `CameraHandle`
- Les dimensions sont déjà retournées par le hook `useCameraPreview` (dans le champ `dimensions`), il suffit à la Camera de passer ces dimensions à la handle

dlymonk: I kept `dimensions` instead of `streamDimensions` for syntax consistency:

`<HUDComponent
      handle={{ takePicture, error, retry, isLoading, dimensions }}
      cameraPreview={cameraPreview}
    />`
    
  if streamDimensions is more explicit, let me know, I will modify it.

PR #676, opened 2024-02-20
MN-469 Fixing portrait camera + prevent zoom/wide camera

dlymonk:


packages/public/camera-web/src/Camera/hooks/useUserMedia.ts
```diff
@@ -133,6 +133,24 @@ function getStreamDimensions(stream: MediaStream): PixelDimensions {
...
 }
 
+function swapWidthAndHeight(dimensions: PixelDimensions): PixelDimensions {
+  return {
+    width: dimensions.height,
+    height: dimensions.width,
+  };
+}
+
+function isMobileDevice(): boolean {
```
souyahia, line 143: Add this function in the common package, in a util file called `browser.utils.ts` and don't forget to add tests

packages/public/camera-web/src/Camera/hooks/useUserMedia.ts
```diff
@@ -202,10 +220,36 @@ export function useUserMedia(constraints: MediaStreamConstraints): UserMediaResu
...
+              exact: devices
+                .filter(
+                  (device) =>
+                    device.kind === 'videoinput' &&
+                    !device.label.includes('Wide') &&
+                    !device.label.includes('Telephoto') &&
+                    !device.label.includes('Triple') &&
+                    !device.label.includes('Dual') &&
+                    !device.label.includes('Ultra'),
+                )
```
souyahia, line 238: for better readability, I suggest you create a filter function in the file called `isValidCamera`

packages/public/camera-web/src/Camera/hooks/useUserMedia.ts
```diff
@@ -202,10 +220,36 @@ export function useUserMedia(constraints: MediaStreamConstraints): UserMediaResu
...
+                .map((device) => device.deviceId),
+            },
+          };
+          str.getTracks().forEach((track) => track.stop());
```
souyahia, line 242: You can put all of this logic into a function in this file called `getValidCameraDeviceIds` that takes the constraints as parameter and return the array of valid deviceIds. It also stops the stream created etc.

Then in the main function you just simply update the media constraints with the deviceIds obtained from this function and fetch the stream

packages/public/camera-web/src/Camera/hooks/useUserMedia.ts
souyahia, line 238: For better readability, extract this arrow function into a normal function in the file called `isValidCamera` that takes the device as an argument and returns a boolean indicating if the camera is valid

souyahia:
Also now that I think about it : when I wrote the code for the Camera in the refacto, I thought it was a good idea to let the user select the device manually using a deviceId in the constraints given to the Camera component. But after Deepak's work on the camera, it appears that this is a mistake : we should always handle the device selection ourselves to prevent users from using zoomed cameras by inadvertance.

So could you take the time to remove this option in the code please? Remove every option to select the deviceId in the camera constraints (everywhere in the code : in the Camera props, in the getMediaConstraints, in the type definitions, in the doc etc.)

packages/public/camera-web/src/Camera/hooks/useUserMedia.ts
```diff
@@ -215,5 +259,22 @@ export function useUserMedia(constraints: MediaStreamConstraints): UserMediaResu
...
+    const portrait = window.matchMedia('(orientation: portrait)');
+
+    const handleOrientationChange = () => {
+      if (stream) {
+        const dimensionsStr = getStreamDimensions(stream);
+        setDimensions(isMobileDevice() ? swapWidthAndHeight(dimensionsStr) : dimensionsStr);
+      }
+    };
+    portrait.addEventListener('change', handleOrientationChange);
+    // console.log(portrait);
```
souyahia, line 272: renove this

souyahia-monk (approved):

PR #679, opened 2024-02-23
Retry functionality when HTTP request fails for TASK

rohitsharma120582:
## Technical description
Implement retry logic when HTTP request fails for damage-detection TASKs.


packages/corejs/src/tasks/index.ts
```diff
@@ -82,12 +86,58 @@ export async function updateOne(
...
+        console.error('Status:', error.response.status);
+        console.error('Headers:', error.response.headers);
+      } else if (error.request) {
+        // The request was made but no response was received
+        console.error('Request:', error.request);
+      } else {
+        // Something happened in setting up the request that triggered an Error
+        console.error('Error message:', error.message);
+      }
+      console.error('Config:', error.config);
```
souyahia, line 113: Remove all of these console logs

PR #680, opened 2024-02-23
MN-487 Implemented wireframes to sights package

dlymonk:


packages/public/sights/research/data/wireframes360/wireframes360.json
```diff
@@ -0,0 +1,254 @@
+{
+  "fesc20-front-left": {
+    "id": "fesc20-front-left",
+    "category": "exterior",
+    "label": "front-left",
+    "overlay": "fesc20-front-left.svg",
+    "vehicle": "fesc20",
+    "mirror_sight": "fesc20-front-right",
+    "tasks": ["damage_detection"]
```
souyahia-monk, line 9: Hmm for me the Wireframe360 are not sights, so you should not have the exact same format as the sights for the JSON. Instead, you could have something like this :

```json
{
  "fesc20-front-left": {
    "id": "fesc20-front-left",
    "label": "front-left",
    "overlay": "fesc20-front-left.svg",
    "vehicle": "fesc20",
    "viewpoint": "front-left"
  }
}
```

This means that you need to update the schema for the JSONs so that it validates this data structure. The viewpoint property is either : front-left / front-right / rear-left / rear-right

souyahia-monk:
Ok so I think there was a misunderstanding about how to implement the wireframes in the refacto :

- The wireframes360 are not sights, and are completely unrelated to sights. They are other types of objects exported by the sights package
- This means that they should have :
  - Their own location in the research directory
  - Their own data structure and JSON schema for validating this structure
  - Their own exported object in the sights package

A good data structure for the car wireframes could look like that :

```json
{
  "fesc20-front-left": {
    "id": "fesc20-front-left",
    "label": "front-left",
    "overlay": "fesc20-front-left.svg",
    "vehicle": "fesc20",
    "mirror_sight": "fesc20-front-right",
    "tasks": ["damage_detection"],
    "viewpoint": "front-left"
  }
}
```

With `viewpoint` being an enumeration with either `front-left`, `front-eight`, `rear-left` or `rear-right`.

And then, in the building code that generates the JSONs, you should have a new mapper function called something like `mapWireframes360` or something like that, that processes only the wireframes360 data, and leave the "sight" code untouched.

Finally, the sights package should export the wireframes360 in a dedicated object like this :

```typescript
import { wireframes360 } from '@monkvision/sights`;

const wireframe = wireframes360['fesc20-front-right'];
```

Oh and also don't forget to :
- Declare types for the Wireframe360 objects in the types packages
- Write TSDoc for types and objects exported
- Write documentation in the README

souyahia-monk:
Closing this PR for now because the branches are going to be modified in the main repo. We will re-open a new one for this feature when the time comes.

PR #778, opened 2024-06-06
💄 make login error message look good on vertical

arunachalam-monk:
## Overview
Jira Ticket Reference : [MN-547](https://acvauctions.atlassian.net/browse/MN-547)
Change login Error message appearance

[MN-547]: https://acvauctions.atlassian.net/browse/MN-547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/common-ui-web/src/components/Login/Login.styles.ts
+++ b/packages/common-ui-web/src/components/Login/Login.styles.ts
@@ -7,7 +7,7 @@ export const styles: Styles = {
     display: 'flex',
     flexDirection: 'column',
     justifyContent: 'center',
-    gap: '1.5em',
+    gap: '20px',
     alignItems: 'center',
     textAlign: 'center',
     maxWidth: '70%',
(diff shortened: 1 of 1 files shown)


packages/common-ui-web/src/components/Login/Login.styles.ts
```diff
@@ -7,7 +7,11 @@ export const styles: Styles = {
     display: 'flex',
     flexDirection: 'column',
     justifyContent: 'center',
+    gap: '1.5em',
```
souyahia-monk, line 10: Use pixels instead of `em`, you're right in the fact that `em` are better for responsiveness, but unfortunately, it can lead to unexpected behavior on certain apps for our clients

packages/common-ui-web/src/components/Login/Login.styles.ts
```diff
@@ -7,7 +7,11 @@ export const styles: Styles = {
...
     alignItems: 'center',
+    textAlign: 'center',
+    maxWidth: '70%',
+    margin: 'auto',
```
souyahia-monk, line 14: don't use margin auto

PR #783, opened 2024-06-11
🐛 svg styles override unintentionally

arunachalam-monk:
## Overview
Jira Ticket Reference : [MN-554](https://acvauctions.atlassian.net/browse/MN-554)
seems `getAttributes` is used to add styles it unintentionally overrides the style.

[MN-554]: https://acvauctions.atlassian.net/browse/MN-554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/documentation/src/components/Sights/SightCard/SightCard.tsx
+++ b/documentation/src/components/Sights/SightCard/SightCard.tsx
@@ -5,6 +5,7 @@ import { Sight, VehicleDetails, VehicleModel } from '@monkvision/types';
 import { CopyPopup, CopyPopupHandle } from '@site/src/components';
 import { DynamicSVG } from '../../domOnly';
 import styles from './SightCard.module.css';
+import { DynamicSVGCustomizationFunctions } from '@monkvision/common-ui-web';
 
 const vehicleModelDisplayOverlays: Record<Exclude<VehicleModel, VehicleModel.ALL>, string> = {
   [VehicleModel.AUDIA7]: sights['audia7-WKJlxkiF'].overlay,
@@ -53,11 +54,10 @@ export function SightCard({ item }: SightCardProps) {
     copyPopupRef.current?.open();
   };
 
-  const getOverlayAttributes = () => ({
-    style: {
-      stroke: isDarkTheme ? '#ffffff' : '#000000',
-    },
-  });
+  const getOverlayAttributes: DynamicSVGCustomizationFunctions['getAttributes'] = (element) => {
+    element.attributeStyleMap.set('stroke', isDarkTheme ? '#ffffff' : '#000000');
+    return null;
+  };
 
   return (
     <>
--- a/documentation/src/components/Sights/SightCard/SightCard.tsx
+++ b/documentation/src/components/Sights/SightCard/SightCard.tsx
@@ -54,13 +54,11 @@ export function SightCard({ item }: SightCardProps) {
     copyPopupRef.current?.open();
   };
 
-  const getOverlayAttributes: DynamicSVGCustomizationFunctions['getAttributes'] = () => {
-    return {
-      style: {
-        stroke: isDarkTheme ? '#ffffff' : '#000000',
-      },
-    };
-  };
+  const getOverlayAttributes: DynamicSVGCustomizationFunctions['getAttributes'] = () => ({
+    style: {
+      stroke: isDarkTheme ? '#ffffff' : '#000000',
+    },
+  });
 
   return (
     <>
--- a/documentation/src/components/Sights/SightCard/SightCard.tsx
+++ b/documentation/src/components/Sights/SightCard/SightCard.tsx
@@ -3,9 +3,9 @@ import { useColorMode } from '@docusaurus/theme-common';
 import { labels, sights } from '@monkvision/sights';
 import { Sight, VehicleDetails, VehicleModel } from '@monkvision/types';
 import { CopyPopup, CopyPopupHandle } from '@site/src/components';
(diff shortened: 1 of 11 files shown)


documentation/src/components/Sights/SightCard/SightCard.tsx
```diff
@@ -53,11 +54,13 @@ export function SightCard({ item }: SightCardProps) {
     copyPopupRef.current?.open();
   };
 
-  const getOverlayAttributes = () => ({
-    style: {
-      stroke: isDarkTheme ? '#ffffff' : '#000000',
-    },
-  });
+  const getOverlayAttributes: DynamicSVGCustomizationFunctions['getAttributes'] = () => {
+    return {
```
souyahia-monk, line 58: replace `return` statement with direct arrow function return

documentation/src/components/Sights/TopBar/TopBar.tsx
```diff
@@ -2,13 +2,13 @@ import { labels, sights, vehicles } from '@monkvision/sights';
...
 import clsx from 'clsx';
-import React, { ReactElement, useRef } from 'react';
+import React, { FunctionComponentElement, useRef } from 'react';
 import styles from './TopBar.module.css';
 
 export interface ListItem {
   id: string;
   lookups: string[];
-  component: ReactElement;
+  component: FunctionComponentElement<typeof SightCard>;
```
souyahia-monk, line 11: Why did you change this exactly? I think `component` could be any type of displayable element. Not ONLY a `SightCard` component

souyahia-monk:
Also : you forgot to add a test case for this PR

documentation/src/components/Sights/TopBar/TopBar.tsx
arunachalam-monk: Technically it can be any component we wanted. but we are only using SightCard from `pages/Sights/index.tsx` its not clear. We need to go and look into other files to find what are the components we are using. 
Now we can see that in index.tsx file itself we can see the component using

souyahia-monk: I think we just need to think about the objective of the ListItem object. What is its purpose? It is just an item in the list that contains a `component` property that is any React element to be displayed on the screen that's it. So let's just make it like that. I see no advantage of strongly typing the `component` property

packages/common-ui-web/src/components/DynamicSVG/hooks/useJSXTransformAttributes.ts
```diff
@@ -42,8 +42,8 @@ export function useJSXTransformAttributes(element: Element) {
         return {
           ...prev,
           [key]: value,
-        };
-      }, {}),
+        } as Partial<JSX.IntrinsicElements>;
+      }, {}) as JSX.IntrinsicElements,
```
souyahia-monk, line 46: Remove this type cast and instead add a generic type to reduce :

```typescript
    return useMemo(
    () =>
      element.getAttributeNames().reduce<Partial<JSX.IntrinsicElements>>((prev, attr) => {
        const { key, value } = tranformJsxAttribute({
          key: attr,
          value: element.getAttribute(attr),
        });
        return {
          ...prev,
          [key]: value,
        };
      }, {}),
    [element],
  );
```

documentation/src/components/Sights/TopBar/TopBar.tsx
```diff
@@ -21,8 +21,7 @@ export interface TopBarProps {
   onSelectTab: (tab: Tab) => void;
   onLookupInput: (value: string) => void;
 }
-
-export const tabs = {
+export const tabs: Record<'sights' | 'vehicles' | 'labels', Tab> = {
```
souyahia-monk, line 24: Remove this type declaration : let TypeScript infer the type

packages/common-ui-web/test/components/DynamicSVG/hooks/useCustomAttributes.test.ts
```diff
@@ -7,50 +7,63 @@ describe('useCustomAttributes hook', () => {
...
     const { result, unmount } = renderHook(useCustomAttributes, {
-      initialProps: { element, groupIds, getAttributes },
+      initialProps: { element, groupIds, getAttributes, style: customAttr.style },
     });
 
     expect(getAttributes).toHaveBeenCalledWith(element, groupIds);
     expect(result.current).toEqual(customAttr);
     unmount();
   });
+  it('should apply style correctly', () => {
```
souyahia-monk, line 56: blank line above `it` statement

documentation/src/components/Sights/TopBar/TopBar.tsx
arunachalam-monk: It's inferring the types only while we use the tabs value not putting any restrictions while writing the code.
for example.
```ts
const tabs = {/** the tab value is wrong */} // won't tell error on hear
const a: Tab = tabs.sight; // will tell error.
```

The main problem is if someone adds a new property for one of the tabs and uses it we won't be able to find it on code review. after 10 iterations on the tabs value the things will be out of track.

packages/common-ui-web/test/components/DynamicSVG/hooks/useCustomAttributes.test.ts
souyahia-monk: @arunachalam-monk this is not fixed

PR #787, opened 2024-06-13
Fixed analytics + added monitoring & analytics state

dlymonk:
## Overview

- Added analytics `setUserId` when inspectionId is passed as URL params.
- Added analytics and monitoring app state in common-ui-web component: `selectVehicleType`, `login`, `createInspection`.

### DIFF
--- a/packages/inspection-capture-web/src/PhotoCapture/PhotoCapture.tsx
+++ b/packages/inspection-capture-web/src/PhotoCapture/PhotoCapture.tsx
@@ -35,6 +35,7 @@ import {
   useUploadQueue,
   useBadConnectionWarning,
   useAdaptiveCameraConfig,
+  useTracking,
 } from './hooks';
 import { PhotoCaptureHUD, PhotoCaptureHUDProps } from './PhotoCaptureHUD';
 import { styles } from './PhotoCapture.styles';
@@ -133,12 +134,13 @@ export function PhotoCapture({
     customComplianceThresholdsPerSight,
   });
   const { t } = useTranslation();
-  const { handleError } = useMonitoring();
+  const monitoring = useMonitoring();
   const [currentScreen, setCurrentScreen] = useState(PhotoCaptureScreen.CAMERA);
   const dimensions = useWindowDimensions();
   const analytics = useAnalytics();
   const loading = useLoadingState();
   const addDamageHandle = useAddDamageMode();
+  useTracking({ inspectionId, authToken: apiConfig.authToken });
   useComplianceAnalytics({ inspectionId, sights });
   const { adaptiveCameraConfig, uploadEventHandlers: adaptiveUploadEventHandlers } =
     useAdaptiveCameraConfig({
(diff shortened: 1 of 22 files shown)


packages/inspection-capture-web/src/PhotoCapture/PhotoCapture.tsx
```diff
@@ -123,6 +126,7 @@ export function PhotoCapture({
   ...initialCameraConfig
 }: PhotoCaptureProps) {
   useI18nSync(lang);
+  const { authToken, vehicleType, steeringWheel } = useMonkAppState();
```
souyahia-monk, line 129: I don't think we should use `useMonkAppState` in the `PhotoCapture` component :
- If the `vehicleType`, `steeringWheel` etc are defined in the `useMonkAppState`, it means that they already are set in the monitoring and analytics (thanks to `useAppStateMonitoring` and `useAppStateAnalytics`).
- I asked you to also set these value in the PhotoCapture component in the case that the user is specifically NOT using the `useMonkAppState` hook, but still using the `PhotoCapture` component.
- Of course the `PhotoCapture` component does not know the vehicle type and steering wheel position and it's fine, but at least it knows the inspection ID and the auth token.

packages/inspection-capture-web/src/PhotoCapture/PhotoCapture.tsx
```diff
@@ -140,6 +144,8 @@ export function PhotoCapture({
   const loading = useLoadingState();
   const addDamageHandle = useAddDamageMode();
   useComplianceAnalytics({ inspectionId, sights });
+  useAppStateMonitoring({ authToken, inspectionId, vehicleType, steeringWheel });
+  useAppStateAnalytics({ inspectionId });
```
souyahia-monk, line 148: And I think these two hooks should not be exported and only be used inside the `common` package

packages/inspection-capture-web/src/PhotoCapture/hooks/usePhotoCaptureSightState.ts
```diff
@@ -191,10 +191,11 @@ export function usePhotoCaptureSightState({
         alreadyTakenSights: alreadyTakenSights.length,
         totalSights: captureSights.length,
         sightSelected: notCapturedSights.length > 0 ? notCapturedSights[0].label : null,
+        captureCompleted: notCapturedSights.length === 0,
```
souyahia-monk, line 194: Shouldn't we also update the `captureCompleted` property after each sight captured and not only when fetching the inspection?

dlymonkai: We handle this part in this callback: 
``` js
const handleGalleryValidate = () => {
    startTasks()
      .then(() => {
        analytics.trackEvent('Capture Completed');
        analytics.setUserProperties({ captureCompleted: true });
        onComplete?.();
      })
      .catch((err) => {
        loading.onError(err);
        handleError(err);
      });
  };
  ```

PR #797, opened 2024-06-20
Feat/mn 556/alert for goback browser

arunachalam-monk:
## Overview

Jira Ticket Reference : [MN-556](https://acvauctions.atlassian.net/browse/MN-556)

When redirecting outside of the page it should ask for confirmation.

[MN-556]: https://acvauctions.atlassian.net/browse/MN-556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/demo-app/src/components/AppRouter.tsx
```diff
@@ -10,6 +11,7 @@ import {
 import { App } from './App';
 
 export function AppRouter() {
+  usePreventExit();
```
souyahia-monk, line 14: Place this in the `PhotoCapture` component instead

packages/common/src/hooks/usePreventExit.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect } from 'react';
+
+/**
+ * Hook that prevents the user from leaving the page without confirmation.
+ * @example
```
souyahia-monk, line 5: blank line above `@example`

packages/common/src/hooks/usePreventExit.ts
```diff
@@ -0,0 +1,24 @@
...
+ * function MyComponent() {
+ *  usePreventExit();
+ *  return <div>My Component</div>;
+ * }
+ * ```
+ */
+export function usePreventExit() {
+  useEffect(() => {
+    window.onbeforeunload = () => {
+      return 'Are you sure you want to leave?';
```
souyahia-monk, line 18: Instead of displaying a hard coded message :

- Make it so that the component accepts a single prop called `confirmationMessage` which is an optional string (default value : 'Are you sure you want to leave?')
- Then, in the `PhotoCapture`, pass a translated message like this : 

```tsx
usePreventExit(t('exit-message'));
```

packages/common/test/hooks/usePreventExit.test.ts
```diff
@@ -0,0 +1,18 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { usePreventExit } from '../../src';
+
+describe('usePreventExit hook', () => {
+  it('should set and unset while component being mount and unmount', () => {
```
souyahia-monk, line 5: *It should properly set and unset the onbeforeunload event handler*

packages/common/src/hooks/usePreventExit.ts
arunachalam-monk: actually whatever messages we have on alert it always says "you have some unsaved changes" not out messages

souyahia-monk: Okay so what is the point of having this hardcoded string? Can't we just remove it?

arunachalam-monk: It should just have some truthy value to show some of the alerts. I have seen mentioned somewhere like on legacy browsers it can will show but now a days modern browsers it's not showing. Probably due to some vulnerability.

arunachalam-monk: https://stackoverflow.com/questions/45088861/whats-the-point-of-beforeunload-returnvalue-if-the-message-does-not-get-set

PR #813, opened 2024-07-17
Added additional config validator for thumbnail domain + fetch thumbnail when addImage

dlymonkai:


documentation/src/utils/schemas.ts
```diff
@@ -212,6 +212,32 @@ export const CreateInspectionDiscriminatedUnionSchema = z.discriminatedUnion(
...
 );
 
+const apiDomain = ['api.monk.ai/v1', 'api.preview.monk.ai/v1', 'api.staging.monk.ai/v1'] as const;
+const thumbnailDomain = [
+  'europe-west1-monk-staging-321715.cloudfunctions.net/image_resize',
+  'europe-west1-monk-preview-321715.cloudfunctions.net/image_resize',
+  'europe-west1-monk-prod.cloudfunctions.net/image_resize',
+] as const;
+
+export const DomainSchema = z
```
souyahia-monk, line 222: Rename this to `DomainsSchema`

packages/network/src/api/image/requests.ts
```diff
@@ -63,6 +63,10 @@ export interface AddBeautyShotImageOptions {
    * The list of tasks to run for this image.
    */
   tasks: TaskName[];
+  /**
+   * Boolean indicating if a thumbnail request will be send when addImage is called.
+   */
+  enableThumbnail: boolean;
```
souyahia-monk, line 69: _...will be sen**t** when..._

Also : rename variable to `useThumbnailCaching`, set it to optional with default : `true`

packages/network/src/api/image/requests.ts
```diff
@@ -96,6 +100,10 @@ export interface Add2ShotCloseUpImageOptions {
    * Boolean indicating if this picture is the first picture or the second picture of the 2-shot process.
    */
   firstShot: boolean;
+  /**
+   * Boolean indicating if a thumbnail request will be send when addImage is called.
+   */
+  enableThumbnail: boolean;
```
souyahia-monk, line 106: _...will be sen**t** when..._

Also : rename variable to `useThumbnailCaching`, set it to optional with default : `true`

packages/network/src/api/image/requests.ts
```diff
@@ -323,6 +332,12 @@ function createLocalImage(options: AddImageOptions): Image {
   return image;
 }
 
+function fetchThumbnail(options: AddImageOptions, image: Image): void {
+  if (options.uploadType !== ImageUploadType.VIDEO_FRAME && options.enableThumbnail) {
+    ky.get(image.thumbnailPath);
+  }
+}
+
```
souyahia-monk, line 340: Rename function to `precacheThumbnail`. Also, this funciton should return a promise like this :

```javascript
return options.uploadType !== ImageUploadType.VIDEO_FRAME && options.enableThumbnail
  ? ky.get(image.thumbnailPath).then(() => {})
  : Promise.resolve();
```

packages/network/src/api/image/requests.ts
```diff
@@ -365,8 +380,10 @@ export async function addImage(
     const image = mapApiImage(
       body,
       options.inspectionId,
+      config.thumbnailDomain,
       (options as AddBeautyShotImageOptions).compliance,
     );
+    fetchThumbnail(options, image);
```
souyahia-monk, line 386: Properly handle the error case for this promise call (we try to avoid unhandled promise rejections) :

fetchThumbnail(options, image).catch(console.error);

packages/network/src/api/utils.ts
```diff
@@ -0,0 +1,30 @@
+import { THUMBNAIL_RESIZE_RATIO, Image } from '@monkvision/types';
```
souyahia-monk, line 1: Why is the `THUMBNAIL_RESIZE_RATIO` constant declared in the types package by the way? If it is only used by the network package, I think we should put it in the network package.

packages/network/src/api/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+function getResizedDimension(originalDimension: number): number {
+  return Math.round(originalDimension * THUMBNAIL_RESIZE_RATIO);
+}
+
+/**
+ * Params of getThumbnailUrl function.
+ */
+export type GetThumbnailUrlParams = {
+  thumbnailDomain: string;
+} & ({ image: Image } | Pick<Image, 'path' | 'width' | 'height'>);
```
souyahia-monk, line 12: Usually, when you have a highly modular type like that for the parameters of your function, it means that your function is adapting to its usage across the project. Sometimes it is intended. But for small standalone util functions like that, I think it should be the other way around :

- You have a small standalone util function that does one specific job (here it is returning the thumbnail URL for a given image)
- Simply think about what this function needs to return the expected result. Here, it just needs the thumbnail domain, the path, width and height of the image and that's it
- This gives you the signature of the function :

```typescript
export function getThumbnailUrl(
  thumbnailDomain: string,
  path: string,
  dimensions: PixelDimensions,
): string {
  ...
}
```

Then, in your project, the different components that use your util function will have to adapt to the given parameters. Of course this is not a perfect rule that we have to apply every time, but for small standalone functions like these I think it's easier to understand.

souyahia-monk (changes requested):

packages/network/src/api/image/requests.ts
```diff
@@ -323,6 +332,12 @@ function createLocalImage(options: AddImageOptions): Image {
   return image;
 }
 
+function precacheThumbnail(options: AddImageOptions, image: Image): Promise<void> {
+  return options.uploadType !== ImageUploadType.VIDEO_FRAME && options.useThumbnailCaching !== false
+    ? ky.get(image.thumbnailPath).then(() => {})
+    : Promise.resolve();
```
souyahia-monk, line 338: I think the `if` statement to check if the request should be made should not be done in this function. This function is called `precacheThumbnail` so it should just do what it is supposed to do : make a precache request.

The conditions to check if the request should be made or not should be in the add image request.

souyahia-monk (approved):

PR #990, opened 2026-03-27
Allow vehicle walkaround in both directions

gabrieltmonkai:
## Overview

Jira Ticket Reference : [MN-834](https://acvauctions.atlassian.net/browse/MN-XXX)

Implemented vehicle walkaround support in both directions based on segments

[MN-834]: https://acvauctions.atlassian.net/browse/MN-834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/common-ui-web/src/components/VehicleWalkaroundIndicator/useVehicleWalkaroundIndicatorState.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Handle returned by the useVehicleWalkaroundIndicatorState hook.
+ */
+export interface VehicleWalkaroundIndicatorStateHandle {
+  /**
+   * Array of covered segments for visual display.
+   */
+  coveredSegments: CoveredSegment[];
+}
+
+const DEGREE_GRANULARITY = 5;
```
dlymonkai: Could you move this to the top of the file, right after the imports? By convention, constants should be declared at the beginning.

packages/common-ui-web/src/components/VehicleWalkaroundIndicator/VehicleWalkaroundIndicator.types.ts
```diff
@@ -1,15 +1,30 @@
...
  */
 export interface VehicleWalkaroundIndicatorProps {
   /**
    * The rotation of the user around the vehicle (0-360 degrees).
+   * This is calculated from the device orientation alpha and the starting position.
    * - 0: POV in the front of the car
    * - 180: POV is behind the back of the car
    */
-  alpha: number;
+  walkaroundPosition: number;
```
dlymonkai: This prop rename could be a breaking change for our clients, so we should keep the current name.

packages/common-ui-web/src/components/VehicleWalkaroundIndicator/VehicleWalkaroundIndicator.types.ts
```diff
@@ -38,4 +53,11 @@ export interface VehicleWalkaroundIndicatorProps {
    * Boolean indicating if the progress bar should be visible (Circle and vehicle).
    */
   showProgressBar?: boolean;
+  /**
+   * Boolean indicating if segment tracking is active.
+   * When true, the component will track and display covered segments.
+   *
+   * @default false
+   */
```
dlymonkai, line 63: I’m not sure about the isTracking prop. Could you use showProgressBar instead?

packages/inspection-capture-web/src/VideoCapture/hooks/useVehicleWalkaround.ts
```diff
@@ -23,6 +23,35 @@ export interface VehicleWalkaroundHandle {
    * The current position of the user around the vehicle (between 0 and 360).
    */
   walkaroundPosition: number;
+  /**
+   * Percentage of the walkaround completed.
+   */
+  coveragePercentage: number;
+}
+
+const DEGREE_GRANULARITY = 5;
```
dlymonkai: This value is declared twice in `VideoCapture` and `VehicleWalkaroundIndicator`. Could you share this value somewhere ? Is `DEGREE_GRANULARITY` really necessary?

packages/inspection-capture-web/src/VideoCapture/hooks/useVehicleWalkaround.ts
```diff
@@ -23,6 +23,35 @@ export interface VehicleWalkaroundHandle {
...
+const TOTAL_SEGMENTS = 360 / DEGREE_GRANULARITY;
+const SMOOTHING_FACTOR = 0.3;
+const STABILIZATION_THRESHOLD_DEGREES = 30;
+const STABILIZATION_FRAMES = 3;
+
+function normalizeAngle(angle: number): number {
+  const normalized = angle % 360;
+  return normalized < 0 ? normalized + 360 : normalized;
```
dlymonkai: There is the same function in `VehicleWalkaroundIndicator`, could you move this function in the `common package` then import it here?

packages/inspection-capture-web/src/VideoCapture/hooks/useVehicleWalkaround.ts
```diff
@@ -32,34 +61,94 @@ export function useVehicleWalkaround({
   alpha,
 }: UseVehicleWalkaroundParams): VehicleWalkaroundHandle {
   const [startingAlpha, setStartingAlpha] = useState<number | null>(null);
-  const [checkpoint, setCheckpoint] = useState(45);
-  const [nextCheckpoint, setNextCheckpoint] = useState(90);
+  const [coveredSegments, setCoveredSegments] = useState<Set<number>>(new Set());
+  const smoothedAlphaRef = useRef<number | null>(null);
+  const prevStartingAlphaRef = useRef<number | null>(null);
```
dlymonkai: I’m not sure about the purpose of all these refs. Why do we need a ref for alpha? We already have the startingAlpha state.

packages/inspection-capture-web/src/VideoCapture/hooks/useVehicleWalkaround.ts
```diff
@@ -32,34 +61,94 @@ export function useVehicleWalkaround({
   alpha,
 }: UseVehicleWalkaroundParams): VehicleWalkaroundHandle {
   const [startingAlpha, setStartingAlpha] = useState<number | null>(null);
```
dlymonkai, line 55: What do you think about having this state updated if the alpha go through the startingalpha in the opposite side (walking back).

dlymonkai, line 55: What do you think about having this state updated if the alpha go through the startingalpha in the opposite side (walking back).

packages/inspection-capture-web/src/VideoCapture/hooks/useVehicleWalkaround.ts
gabrieltmonkai: Right, I've left it in `useVehicleWalkaroundIndicatorState` and exported it.

It's necessary because `useVehicleWalkaround` displays the visual indicator of coverage and `VehicleWalkaroundIndicator` tracks the actual position and calculates which segments are covered.

If they use different granularity values, the visual display would be incorrect.

packages/inspection-capture-web/src/VideoCapture/hooks/useVehicleWalkaround.ts
gabrieltmonkai: That's a good question, I tried to get rid of them, but couldn't figure out another way to solve those 2 out of 10 cases when at the beginning of video recording, the `alpha` started offset or `coveredSegments` was `2` or `3` already.

We're using `prevStartingAlphaRef` to know whether or not we've had previous covered segments. It's the rare case when at resume or retaking the video after discard, we would have the alpha slightly offset, plus a couple of segments already covered.

Even though we reset the `alpha`, this is extra safety to do a "hard reset" if the user will fall in those 2 out of 10 cases when alpha starts offset.

For the other refs, this is just a non-rendering pattern when we want to do calculations that reflect in real-time, without triggering a render, but while the hook is heavy triggered by the nature of often the alpha updates once the phone is moving.

They are more specific to the beginning of the video tracking after `resume` or `discard video`, to prevent those 2 issues:
- `alpha` starting offset
- `coveredSegments` starting with a value between 2-4 by default

packages/inspection-capture-web/src/VideoCapture/hooks/useVehicleWalkaround.ts
gabrieltmonkai: I'm not sure about the direction, do you point towards tracking the covered segments to the left/right based on `startingAlpha`?

packages/inspection-capture-web/src/VideoCapture/hooks/useVehicleWalkaround.ts
dlymonkai: To be honest, I’m not a fan of using these four refs. It feels like an anti-pattern.

If the alpha is offset or segments are covered, it means something isn’t working properly in the code. This hook likely needs a full rework.

packages/inspection-capture-web/src/VideoCapture/hooks/useVehicleWalkaround.ts
dlymonkai: Yes this will be updated if alpha is going clockwise.

packages/common-ui-web/src/components/VehicleWalkaroundIndicator/useVehicleWalkaroundIndicatorState.ts
```diff
@@ -0,0 +1,100 @@
+import { useEffect, useRef, useState } from 'react';
+import { angleToSegment, segmentToAngle } from '@monkvision/common';
+import { CoveredSegment } from './VehicleWalkaroundIndicator.types';
+
+/**
+ * Granularity (in degrees) used for dividing the 360-degree circle into segments for vehicle walkaround tracking.
+ * This constant determines the precision of position tracking during vehicle inspections.
+ */
+export const DEGREE_GRANULARITY = 5;
```
dlymonkai: This constant is defined in `common-ui-web` but imported in `inspection-capture-web`. This creates an inverted dependency, it should not depend on UI packages.

Is this const really necessary? The original implementation had not granularity.

If really needed, it should be passed as a prop to the component

packages/common-ui-web/src/components/VehicleWalkaroundIndicator/VehicleWalkaroundIndicator.types.ts
```diff
@@ -38,4 +53,29 @@ export interface VehicleWalkaroundIndicatorProps {
...
+  isRecordingPaused?: boolean;
+  /**
+   * Granularity (in degrees) used for dividing the 360-degree circle into segments.
+   *
+   * @default 5
+   */
+  degreeGranularity?: number;
```
dlymonkai: Can't we use a default value of 1. And get rid of `degreeGranularity`

packages/common-ui-web/src/components/VehicleWalkaroundIndicator/VehicleWalkaroundIndicator.types.ts
```diff
@@ -38,4 +53,29 @@ export interface VehicleWalkaroundIndicatorProps {
...
+   * @default false
+   */
+  isRecording?: boolean;
+  /**
+   * Boolean indicating if video recording is paused.
+   * When true, the POV arrow still moves but covered segments are not updated.
+   *
+   * @default false
+   */
+  isRecordingPaused?: boolean;
```
dlymonkai: Could you check the naming and can we remove one of those props?

packages/common-ui-web/src/components/VehicleWalkaroundIndicator/VehicleWalkaroundIndicator.types.ts
```diff
@@ -1,11 +1,25 @@
...
+ * The current state of coverage tracking in the VehicleWalkaroundIndicator.
+ * - `'off'`: No tracking. Covered segments are cleared.
+ * - `'active'`: Tracking is ongoing. Covered segments are updated as the position changes.
+ * - `'paused'`: Tracking is paused. Covered segments are retained but not updated.
+ */
+export enum WalkaroundTrackingState {
+  Off = 'off',
+  Active = 'active',
+  Paused = 'paused',
+}
```
dlymonkai, line 14: By convention, enum must follow this syntax: export enum WalkaroundTrackingState {
  OFF = 'off',
  ACTIVE = 'active',
  PAUSED = 'paused',
}

packages/inspection-capture-web/src/VideoCapture/VideoCaptureHUD/VideoCaptureRecording/VideoCaptureRecording.tsx
```diff
@@ -51,7 +58,7 @@ export function VideoCaptureRecording({
           tooltipPosition={tooltipPosition}
           disabled={recordVideoDisabled}
         />
-        <Button icon='close' variant='outline' />
+        <div style={placeholderTopButton} />
```
gabrieltmonkai: Adapt `VideoCaptureRecordingProps` to how `PhotoCapture` props work:
- have an optional `onClose` property
- have a `showCloseButton`
- conditionally render between Button and div based on the passed props
- update the tests with the new props
- update the global docs with the new props

packages/inspection-capture-web/src/VideoCapture/VideoCaptureHUD/VideoCaptureRecording/VideoCaptureRecording.tsx
```diff
@@ -51,7 +60,17 @@ export function VideoCaptureRecording({
           tooltipPosition={tooltipPosition}
           disabled={recordVideoDisabled}
         />
-        <Button icon='close' variant='outline' />
+
+        {showCloseVideoButton ? (
```
dlymonkai: I think using visibility: hidden is cleaner here since it preserves the layout without adding a placeholder div.

```ts
// This return in the style hook as `closeButton` 
{ visibility: showCloseVideoButton ? 'visible' : 'hidden', }

...

// in Button as props: 
style = { closeButton }
```

packages/inspection-capture-web/src/VideoCapture/VideoCaptureHUD/VideoCaptureRecording/VideoCaptureRecording.tsx
```diff
@@ -51,7 +60,17 @@ export function VideoCaptureRecording({
...
+          <Button
+            icon='close'
+            variant='outline'
+            onClick={onCloseVideo}
+            data-testid='close-video-button'
+          />
+        ) : (
+          <div style={placeholderTopButton} />
```
dlymonkai: So this can be removed from the component and style hook

dlymonkai (approved):

PR #262, opened 2022-05-16
Added device performance measuring, and automatic resolution switch

BenHakimIlyass:
## Technical description

- [x] Ability to changes constraints without directly by changing the constraints
- [ ] Add a device measurements solution, to be tested 

## Tested on

- Chrome
- Safari iOS

## Steps to reproduce

1. Take pictures
2. Weak devices should use FHD instead of QHD

## Screenshots (optional):

No screenshots.

*This Pull Request template has been written and generated by Monk JS repository.*

PR #995, opened 2026-05-21
Added shadow retry for failed upload (timeout & error 5xx)

dlymonkai:
## Overview

Jira Ticket Reference : [MN-850](https://acvauctions.atlassian.net/browse/MN-850)

- [ ] I have updated the unit tests based on the changes I made
- [ ] I have updated the docs (TSDoc / README / global doc) to reflect my changes
- [ ] I have updated the local app configs if needed
- [ ] I have performed self-QA of my feature by testing the apps and packages and made sure that :

[MN-850]: https://acvauctions.atlassian.net/browse/MN-850?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/inspection-capture-web/src/hooks/useUploadQueue.ts
+++ b/packages/inspection-capture-web/src/hooks/useUploadQueue.ts
@@ -11,6 +11,26 @@ import { useRef } from 'react';
 import { useMonitoring } from '@monkvision/monitoring';
 import { CaptureMode } from '../types';
 
+const MAX_RETRY_COUNT = 1;
+
+/**
+ * Checks if an error is retryable (timeout or 5xx server error).
+ */
+export function isRetryableError(err: unknown): boolean {
+  if (err instanceof Error) {
+    if (err.name === 'TimeoutError' || err.message === 'Failed to fetch') {
+      return true;
+    }
+  }
+  if (err && typeof err === 'object' && 'response' in err) {
+    const { response } = err as { response?: { status?: number } };
+    if (response && typeof response.status === 'number' && response.status >= 500) {
+      return true;
+    }
+  }
+  return false;
+}
+
 /**
  * Payload for the onUploadSuccess event handler.
  */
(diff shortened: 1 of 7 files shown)


packages/inspection-capture-web/src/hooks/useUploadQueue.ts
```diff
@@ -11,6 +11,26 @@ import { useRef } from 'react';
 import { useMonitoring } from '@monkvision/monitoring';
 import { CaptureMode } from '../types';
 
+const MAX_RETRY_COUNT = 1;
+
+/**
+ * Checks if an error is retryable (timeout or 5xx server error).
+ */
+export function isRetryableError(err: unknown): boolean {
```
gabrieltmonkai: Do you believe it's a good time to create a `request.utils.ts` in our `common` package for future usage, extendability and simplified declarations as:
- `isTimeoutError` fn 
```typescript
if (err instanceof Error) {
    if (err.name === 'TimeoutError' || err.message === 'Failed to fetch') {
      return true;
    }
  }
``` 

- `isInternalServerError` fn
```typescript
if (err && typeof err === 'object' && 'response' in err) {
    const { response } = err as { response?: { status?: number } };
    if (response && typeof response.status === 'number' && response.status >= 500) {
      return true;
    }
  }
```

- `isRetryableError` fn
```typescript
return isTimeoutError(err) && isInternalServerError(err)
```

And then, we can reuse again the `isTimeoutError` utility function below at line `#262`.

Perhaps we can create a dedicated `RetryableRequest` interface for `{ count: number; retryable: boolean }` as well at line `#221`.

dlymonkai: Took the suggestion, with one tweak: kept the reusable predicates (isTimeoutError, isInternalServerError) in common, but left isRetryableError + RetryableRequest local to useUploadQueue since retry policy could change.
Also FYI — isRetryableError needs ||, not && (with && a plain timeout isn't retryable).

gabrieltmonkai (approved):

PR #997, opened 2026-06-01
VideoCaptureTutorial doesn't load translations

gabrieltmonkai:
## Overview

Jira Ticket Reference : [MN-861](https://acvauctions.atlassian.net/browse/MN-861)

VideoCaptureTutorial now passes `lang` parameter through `VideoCapture` component.

[MN-861]: https://acvauctions.atlassian.net/browse/MN-861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/inspection-capture-web/src/VideoCapture/VideoCaptureTutorial/VideoCaptureTutorial.tsx
+++ b/packages/inspection-capture-web/src/VideoCapture/VideoCaptureTutorial/VideoCaptureTutorial.tsx
@@ -1,5 +1,6 @@
 import { VideoTutorial } from '@monkvision/common-ui-web';
 import { DeviceOrientation } from '@monkvision/types';
+import { useTranslation } from 'react-i18next';
 import { VideoCapturePageLayout } from '../VideoCapturePageLayout';
 
 /**
@@ -14,22 +15,14 @@ export interface VideoCaptureTutorialProps {
    * Use this prop to enforce a specific device orientation for the Camera screen.
    */
   enforceOrientation?: DeviceOrientation;
-  /**
-   * The language to be used by this component.
-   *
-   * @default en
-   */
-  lang?: string | null;
 }
 
 /**
  * This component is a tutorial displayed on top of the camera when the user first starts the video capture.
  */
-export function VideoCaptureTutorial({
-  onClose,
-  enforceOrientation,
-  lang,
-}: VideoCaptureTutorialProps) {
+export function VideoCaptureTutorial({ onClose, enforceOrientation }: VideoCaptureTutorialProps) {
+  const { i18n } = useTranslation();
+
   return (
     <VideoCapturePageLayout
       showLogo={false}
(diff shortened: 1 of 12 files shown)


packages/inspection-capture-web/src/VideoCapture/VideoCaptureTutorial/VideoCaptureTutorial.tsx
```diff
@@ -14,20 +14,34 @@ export interface VideoCaptureTutorialProps {
    * Use this prop to enforce a specific device orientation for the Camera screen.
    */
   enforceOrientation?: DeviceOrientation;
+  /**
+   * The language to be used by this component.
+   *
+   * @default en
+   */
+  lang?: string | null;
```
dlymonkai: Why does lang allow null here? Since there is already a default value (en), string | undefined should be enough

packages/inspection-capture-web/src/VideoCapture/VideoCaptureTutorial/VideoCaptureTutorial.tsx
```diff
@@ -14,20 +14,34 @@ export interface VideoCaptureTutorialProps {
...
       showLogo={false}
       showTitle={false}
       showBackdrop={true}
       showConfirmButton={false}
     >
-      <VideoTutorial orientation={enforceOrientation} onComplete={onClose} />
+      <VideoTutorial
+        orientation={enforceOrientation}
+        onComplete={onClose}
+        lang={lang ?? undefined}
```
dlymonkai: Why are we using the nullish coalescing operator here? If lang should not be null, we could simplify this by updating the prop type instead.

packages/inspection-capture-web/src/VideoCapture/VideoCaptureTutorial/VideoCaptureTutorial.tsx
```diff
@@ -14,20 +14,34 @@ export interface VideoCaptureTutorialProps {
...
 
 /**
  * This component is a tutorial displayed on top of the camera when the user first starts the video capture.
  */
-export function VideoCaptureTutorial({ onClose, enforceOrientation }: VideoCaptureTutorialProps) {
+export function VideoCaptureTutorial({
+  onClose,
+  enforceOrientation,
+  lang,
+}: VideoCaptureTutorialProps) {
```
dlymonkai: Instead of passing lang as a props, it is recommended to use the hook `useTranslation` like this: 
``` ts
const {  i18n } = useTranslation();

...

lang = {i18n.lang}
```

packages/inspection-capture-web/src/VideoCapture/VideoCaptureTutorial/VideoCaptureTutorial.tsx
gabrieltmonkai: The usage was for satisfying `BaseVideoCaptureProps` declaration of lang: ``lang?: string | null;`.

Furthermore, there are 9 other similar declaration in the project and just 5 for `lang?: string;`. 

I suggest addressing all other `| null` references and clean up our code.

dlymonkai (approved):

PR #1002, opened 2026-06-05
Pictures from video are correctly filtered

gabrieltmonkai:
## Overview

Jira Ticket Reference : [MN-864](https://acvauctions.atlassian.net/browse/MN-864)

Pictures from video are correctly filtered when mapping images function is called.

[MN-864]: https://acvauctions.atlassian.net/browse/MN-864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/inspection-capture-web/src/hooks/usePhotoCaptureImages.ts
+++ b/packages/inspection-capture-web/src/hooks/usePhotoCaptureImages.ts
@@ -9,10 +9,7 @@ export function usePhotoCaptureImages(inspectionId: string): Image[] {
   const { state } = useMonkState();
 
   return useMemo(
-    () =>
-      getInspectionImages(inspectionId, state.images, undefined, true).filter(
-        (image) => image.additionalData?.frame_index === undefined,
-      ),
+    () => getInspectionImages(inspectionId, state.images, undefined, true),
     [state.images, inspectionId],
   );
 }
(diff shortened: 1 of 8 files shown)


packages/inspection-capture-web/src/hooks/usePhotoCaptureImages.ts
```diff
@@ -9,7 +9,10 @@ export function usePhotoCaptureImages(inspectionId: string): Image[] {
   const { state } = useMonkState();
 
   return useMemo(
-    () => getInspectionImages(inspectionId, state.images, undefined, true),
+    () =>
+      getInspectionImages(inspectionId, state.images, undefined, true).filter(
+        (image) => image.additionalData?.frame_index === undefined,
```
dlymonkai: @gabrieltmonkai  UsePhotoCaptureImages hook is the shared "all taken images" source, so filtering there changes its behavior for every other consumer. Since only the retake badge actually needs it , I moved the filter to the retakeCount instead. (I have pushed an new commit) 
What do think about it?

gabrieltmonkai: Great catch! Just tested it on hybrid to assure we get the correct retake count when non-compliant photos are taken while in Photo Capture and everything works as expected.

dlymonkai (dismissed):

dlymonkai (approved):

PR #1014, opened 2026-07-23
Implemented adaptive frame selection mechanism

gabrieltmonkai:
## Overview

Jira Ticket Reference : [MN-876](https://acvauctions.atlassian.net/browse/MN-876)

Implemented adaptive frame selection mechanism when moving around the car.

[MN-876]: https://acvauctions.atlassian.net/browse/MN-876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/inspection-capture-web/src/VideoCapture/VideoCaptureHUD/VideoCaptureHUD.tsx
```diff
@@ -77,7 +78,6 @@ export interface VideoCaptureHUDProps
 }
 
 const SCREENSHOT_INTERVAL_MS = 200;
-const FRAME_SELECTION_INTERVAL_MS = 1000;
```
dlymonkai, line 80: If I remember correctly, we talked about adding a new prop for `targetPictureCount`, so  `frameSelectionInterval` can adapt to it.

packages/inspection-capture-web/src/translations/en.json
```diff
@@ -86,6 +86,7 @@
     "recording": {
       "discardDialog": {
         "message": "Do you want to discard the video? You haven' t gone all the way around the vehicle.",
+        "messageMissingFrames": "Do you want to discard the video? Not all of the required photos have been captured yet.",
```
dlymonkai: Could you swap "photos" for "footage" here (and other languages also) — I think we should use another word than "photos" since the user doesn't know we're not actually recording a video (they'd expect "video" and "photos" not to be mixed in the same sentence):

dlymonkai (commented):
test

dlymonkai (commented):
test

dlymonkai (commented):
test l89

dlymonkai (commented):
test l90

packages/inspection-capture-web/src/VideoCapture/hooks/useFrameSelection/useFrameSelection.ts
```diff
@@ -84,7 +92,16 @@ export function useFrameSelection({
...
-  }, frameSelectionInterval);
+  }, [handle, onFrameSelected, handleError]);
+
+  useInterval(flushBestFrame, flushTrigger === undefined ? frameSelectionInterval : null);
+
+  useEffect(() => {
+    if (flushTrigger === undefined) {
+      return;
+    }
+    flushBestFrame();
```
dlymonkai, line 131: **[Bug — CONFIRMED] Stale frame from discarded recording uploaded to the next one (adaptive mode only)**

In adaptive mode `useInterval` is disabled (`null` delay), so `bestFrame.current` is never drained after a discard. When the user starts a new recording, `startSegmentTracking()` immediately increments `flushTrigger`, which triggers this `flushBestFrame()` call — uploading whatever frame was accumulated in the previous (discarded) session into the new recording.

**Reproduction:** record partway → click stop → click "Discard Video" → start a new recording → the first uploaded frame belongs to the old session.

**Fix:** reset `bestFrame.current = null` and `bestScore.current = null` inside `onDiscardDialogDiscardVideo`.

packages/inspection-capture-web/src/VideoCapture/hooks/useSegmentFrameSelection/useSegmentFrameSelection.ts
```diff
@@ -0,0 +1,108 @@
...
+  const [capturedBuckets, setCapturedBuckets] = useState<Set<number>>(new Set());
+
+  useEffect(() => {
+    if (!isRecording) {
+      return;
+    }
+    const currentBucket = Math.floor(walkaroundPosition / bucketSizeDegrees) % totalBuckets;
+    if (!capturedBuckets.has(currentBucket)) {
+      setCapturedBuckets((prev) => new Set(prev).add(currentBucket));
+      setFlushTrigger((value) => value + 1);
```
dlymonkai: **[Bug — CONFIRMED] Last bucket's best frame is always silently dropped**

Flushes fire on *entering* a bucket and upload the *prior* bucket's best frame. When `capturedFramesCount` reaches `targetFramesCount`, the END tooltip appears and the user clicks stop — but no subsequent `flushTrigger` increment ever occurs, so the best frame captured in the final bucket sits in `bestFrame.current` and is discarded when the component unmounts.

**Result:** the API consistently receives `targetFramesCount − 1` real frames (e.g. 39 for a default config of 40). Fixed-rate mode is unaffected because its `useInterval` keeps flushing after `setIsRecording(false)`.

**Fix:** trigger one final flush when `isRecording` transitions to `false` (e.g. watch `isRecording` going false in `useFrameSelection`, or expose a dedicated end-of-recording signal from this hook).

packages/inspection-capture-web/src/VideoCapture/hooks/useSegmentFrameSelection/useSegmentFrameSelection.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }, [walkaroundPosition, isRecording, bucketSizeDegrees, totalBuckets, capturedBuckets]);
+
+  const startSegmentTracking = useCallback(() => {
+    setCapturedBuckets(new Set([0]));
+    setFlushTrigger((value) => value + 1);
```
dlymonkai: **[Bug — CONFIRMED] Initial `startSegmentTracking` flush is always a no-op → every walkaround delivers `targetFramesCount − 1` real uploads**

`startSegmentTracking` increments `flushTrigger` at the moment recording begins, before the screenshot interval has fired even once. The resulting `flushBestFrame()` fires immediately, finds `bestFrame.current === null`, and exits without uploading anything — while bucket 0 is already marked as captured.

Combined with the missing final-bucket flush above, a `targetFramesCount=40` config produces **39 real uploads**, not the "+/−1" stated in the hook's docstring — it is consistently −1.

**Fix:** remove `setFlushTrigger(v => v + 1)` from `startSegmentTracking`. Bucket 0 will be captured naturally once the first screenshot is scored and the user moves to bucket 1.

packages/inspection-capture-web/src/VideoCapture/VideoCaptureHUD/VideoCaptureHUD.tsx
```diff
@@ -80,8 +89,7 @@ export interface VideoCaptureHUDProps
   showCloseVideoButton?: boolean;
```
dlymonkai, line 89: **[Efficiency — PLAUSIBLE] Screenshot interval halved to 100 ms doubles synchronous Laplace load**

`SCREENSHOT_INTERVAL_MS` was changed 200 ms → 100 ms, so `onCaptureVideoFrame` fires 10×/sec instead of 5×/sec. Each call copies raw `ImageData` and enqueues an O(width × height) Laplace convolution on the JS main thread.

At 1080p this is ~20 M pixel ops/sec competing with gyroscope events, React renders, and the upload queue. On mid-range Android this risks saturating the JS thread, causing camera-preview jank and degrading the sharpness of the very frames being scored — the opposite of the intended effect. Consider whether a lower rate (or a device-adaptive rate) is sufficient for the adaptive strategy.

packages/inspection-capture-web/src/VideoCapture/hooks/useSegmentFrameSelection/useSegmentFrameSelection.ts
```diff
@@ -0,0 +1,108 @@
...
+  const totalBuckets = useMemo(
+    () => clampTargetFramesCount(targetFramesCount),
+    [targetFramesCount],
+  );
+  const bucketSizeDegrees = 360 / totalBuckets;
+
+  const [flushTrigger, setFlushTrigger] = useState(0);
+  const [capturedBuckets, setCapturedBuckets] = useState<Set<number>>(new Set());
+
+  useEffect(() => {
```
dlymonkai, line 90: **[Simplification — PLAUSIBLE] Bucket-tracking logic duplicated from `useVehicleWalkaround`**

The `Math.floor(walkaroundPosition / bucketSizeDegrees) % totalBuckets` + `Set<number>` pattern is identical to the one in `useVehicleWalkaround`. The two have already diverged in guard conditions (`useVehicleWalkaround` additionally guards on `startingAlpha !== null`), meaning a future edge-case fix must be applied independently in both.

Consider extracting a shared `useBucketTracker(position, totalBuckets, isActive)` primitive that both hooks consume.

---

**[Efficiency — PLAUSIBLE] `capturedBuckets` in deps double-fires per bucket boundary**

Including `capturedBuckets` in this effect's dependency array causes the body to run twice per new-bucket crossing: once when `walkaroundPosition` changes (guard passes → `setCapturedBuckets + setFlushTrigger`), and again after the state update commits (`capturedBuckets.has()` is now true → early exit). At 60 Hz gyroscope updates this doubles closure invocations at every angular boundary and adds an extra React reconciliation per capture. Consider using a `useRef` for the set to avoid the second run.

dlymonkai (commented):
Automated code review — 3 confirmed correctness bugs and 3 lower-severity findings.

dlymonkai (commented):
test

dlymonkai (commented):
test

dlymonkai (commented):
test

dlymonkai (commented):
test

dlymonkai (commented):
test

dlymonkai (commented):
test

dlymonkai (commented):
test

dlymonkai (commented):
test

dlymonkai (commented):
test

dlymonkai (commented):
test

dlymonkai (commented):
test

packages/inspection-capture-web/src/VideoCapture/hooks/useFrameSelection/useFrameSelection.ts
```diff
@@ -50,18 +69,20 @@ export interface FrameSelectionHandle {
 export function useFrameSelection({
   handle,
   frameSelectionInterval,
+  flushTrigger,
   onFrameSelected,
 }: UseFrameSelectionParams): FrameSelectionHandle {
   const bestScore = useRef<number | null>(null);
   const bestFrame = useRef<ImageData | null>(null);
+  const generation = useRef(0);
+  const [startedFlushes, setStartedFlushes] = useState(0);
```
dlymonkai, line 78: **[Bug — CONFIRMED] Flush counters are never reset after a discard + re-record cycle**

`startedFlushes` and `settledFlushes` accumulate across the entire component lifetime. After a discard the processing screen displays inflated counts.

**Scenario:** First recording: 1 000 queue items + 1 flush → counters reach 1. User discards, re-records 1 000 more → screen shows `totalProcessingFrames=2 001` instead of `1 001`. If `processingQueue.clear()` resets `totalItems` to 0 while these counters stay at 2, the screen instantly shows 100% before the new recording's compression settles.

**Fix:** expose a reset in this hook and call it from `onDiscardDialogDiscardVideo`.

packages/inspection-capture-web/src/VideoCapture/hooks/useFrameSelection/useFrameSelection.ts
```diff
@@ -75,20 +96,46 @@ export function useFrameSelection({
     processingQueue.push(handle.getImageData());
   }, [processingQueue.push]);
```
dlymonkai, line 97: **[Bug — CONFIRMED] `handle` missing from `onCaptureVideoFrame` dependency array**

`onCaptureVideoFrame` captures `handle` via closure but only lists `processingQueue.push` as a dep. If `CameraHandle` is recreated (e.g., after a camera permission error remounts the camera component), this callback calls `getImageData()` on the stale, possibly invalid handle — frames would be from the wrong source or trigger a runtime error.

**Fix:** add `handle` to the dependency array.

packages/inspection-capture-web/src/VideoCapture/hooks/useFrameSelection/useFrameSelection.ts
```diff
@@ -75,20 +96,46 @@ export function useFrameSelection({
...
+  const discardBestFrame = useCallback(() => {
+    generation.current += 1;
     bestScore.current = null;
     bestFrame.current = null;
-  }, frameSelectionInterval);
+  }, []);
+
+  useInterval(flushBestFrame, flushTrigger === undefined ? frameSelectionInterval : null);
+
+  useEffect(() => {
```
dlymonkai, line 127: **[Bug — PLAUSIBLE] `useEffect` fires on initial mount because `flushTrigger=0` is not guarded**

`useSegmentFrameSelection` initialises `flushTrigger` with `useState(0)`. On the first render, `useFrameSelection` receives `0` (not `undefined`), so the guard `if (flushTrigger === undefined) return` passes and `flushBestFrame()` is called immediately.

In practice `bestFrame.current` is `null` at mount so it is a no-op, but in React 18 concurrent mode — where an effect can commit while microtasks from a prior flush are still pending — a Laplace-queue item resolved just before mount could cause a stale frame upload.

**Fix:** `if (!flushTrigger) return` (or explicitly `flushTrigger === undefined || flushTrigger === 0`).

packages/inspection-capture-web/src/VideoCapture/hooks/useFrameSelection/useFrameSelection.ts
```diff
@@ -75,20 +96,46 @@ export function useFrameSelection({
...
+    if (flushTrigger === undefined) {
+      return;
+    }
+    flushBestFrame();
+  }, [flushTrigger]);
```
dlymonkai, line 132: **[Bug — CONFIRMED] `flushBestFrame` missing from `useEffect` dependency array — stale closure**

`flushBestFrame` closes over `handle`, `onFrameSelected`, and `handleError` (line 117). If any of those change after the initial render (e.g., the upload-queue callback is recreated on a network state change), this effect keeps using the original closure and delivers compressed frames to the old callback.

ESLint `react-hooks/exhaustive-deps` would flag this under the airbnb-base rules in `CLAUDE.md`.

**Fix:** add `flushBestFrame` to the dep array (or use a stable `useRef` wrapper).

packages/inspection-capture-web/src/VideoCapture/hooks/useVideoRecording.ts
```diff
@@ -169,10 +198,12 @@ export function useVideoRecording({
...
-      if (
-        getRecordingDurationMs() < minRecordingDuration ||
-        coveragePercentage < MINIMUM_PERCENTAGE_VEHICLE_WALKAROUND_COVERAGE
-      ) {
+      const isDurationTooShort = getRecordingDurationMs() < minRecordingDuration;
+      const isMissingCoverage = isAdaptiveUploadRate
+        ? capturedFramesCount < targetFramesCount
+        : coveragePercentage < MINIMUM_PERCENTAGE_VEHICLE_WALKAROUND_COVERAGE;
+      if (isDurationTooShort || isMissingCoverage) {
+        setIsMissingTargetFrames(isAdaptiveUploadRate && isMissingCoverage);
```
dlymonkai, line 206: **[Bug — CONFIRMED] `isMissingTargetFrames` set to `true` even when the duration is also too short**

When both `isDurationTooShort` and `isMissingCoverage` are true simultaneously, `isMissingTargetFrames` is set to `true`, showing the 'Not all required images captured' dialog — even though the real blocker is the recording duration.

**Scenario:** Adaptive mode, user records for 5 s (< 15 s minimum) and captures only 30/40 frames. Dialog says 'missing images', user walks further. On the next stop the dialog says 'too short' — two-step confusion impossible in fixed-rate mode.

**Fix:**
```ts
setIsMissingTargetFrames(!isDurationTooShort && isAdaptiveUploadRate && isMissingCoverage);
```

packages/inspection-capture-web/src/VideoCapture/VideoCaptureHUD/VideoCaptureHUD.tsx
```diff
@@ -224,7 +273,11 @@ export function VideoCaptureHUD({
             isRecording={isRecording}
             isRecordingPaused={isRecordingPaused}
             coveredSegments={isRecording || isRecordingPaused ? coveredSegments : undefined}
-            isComplete={coveragePercentage >= MINIMUM_PERCENTAGE_VEHICLE_WALKAROUND_COVERAGE}
+            isComplete={
+              isAdaptiveUploadRate
+                ? capturedFramesCount >= effectiveTargetFramesCount
```
dlymonkai, line 278: **[Simplification — PLAUSIBLE] Coverage-complete predicate duplicated at three call sites**

The `isAdaptiveUploadRate ? capturedFramesCount >= effectiveTargetFramesCount : coveragePercentage >= MINIMUM_PERCENTAGE_VEHICLE_WALKAROUND_COVERAGE` condition also appears in `useVideoRecording.ts:202` (stop guard) and `useVideoRecording.ts:274` (tooltip effect) — three independent copies.

A developer who updates two of three creates a silent inconsistency where the button shows 'done' while the stop guard still blocks.

**Fix:** extract `isCaptureComplete(strategy, captured, target, coverage)` and call it at all three sites.

packages/inspection-capture-web/test/VideoCapture/VideoCaptureHUD/VideoCaptureHUD.test.tsx
```diff
@@ -25,13 +25,23 @@ jest.mock('../../../src/VideoCapture/hooks', () => ({
     processedFrames: 986,
     totalProcessingFrames: 6782,
     onCaptureVideoFrame: jest.fn(),
+    flushBestFrame: jest.fn(),
+    discardBestFrame: jest.fn(),
+  })),
+  useSegmentFrameSelection: jest.fn(() => ({
+    flushTrigger: 12,
+    capturedFramesCount: 8,
+    targetFramesCount: 40,
```
dlymonkai, line 34: **[Bug — CONFIRMED] Mock returns `targetFramesCount` but the hook exposes `effectiveTargetFramesCount`**

The `useSegmentFrameSelection` mock uses the key `targetFramesCount:40`, but the hook's actual return shape has `effectiveTargetFramesCount`. The component destructures `effectiveTargetFramesCount`, so it receives `undefined` throughout every test.

This silently breaks three things:
1. `capturedFramesCount >= undefined` is always `false` → `isComplete` is never exercised.
2. `useVideoRecording` is called with `targetFramesCount: undefined` → the adaptive-mode stop guard is never tested.
3. Assertions that check `targetFramesCount` passed to `useVideoRecording` pass only because both sides are `undefined` — false confidence.

**Fix:** rename the mock key to `effectiveTargetFramesCount`.

dlymonkai (commented):
**[Efficiency — PLAUSIBLE] `laplaceScores.ts:44` — Single-pass variance has catastrophic cancellation on low-contrast frames**

The `E[X²] − E[X]²` single-pass formula (`sumOfSquares / count - mean * mean`) has catastrophic cancellation for nearly-uniform (low-contrast) images. On a foggy or overcast scene, `sumOfSquares/count` and `mean*mean` are both small positives whose difference is a third-order quantity computed in floating-point — the result can be slightly negative (caught by `Math.max`) or epsilon-positive, so two adjacent frames with different sharpness both report `std≈0` and the first frame 'wins' by default. A two-pass implementation (compute mean first, then sum squared deviations) avoids this entirely.

packages/inspection-capture-web/src/VideoCapture/hooks/useFrameSelection/useFrameSelection.ts
```diff
@@ -73,22 +103,60 @@ export function useFrameSelection({
...
+    setStartedFlushes((count) => count + 1);
+    handle
+      .compressImage(frame)
+      .then((picture) => {
+        if (flushGeneration === generation.current) {
+          onFrameSelected?.(picture);
+        }
+      })
+      .catch(handleError)
+      .finally(() => setSettledFlushes((count) => count + 1));
```
dlymonkai, line 125: **[Bug — CONFIRMED] `settledFlushes` not guarded by generation — `resetProcessingCounters` can't prevent a concurrent in-flight flush from corrupting the counter**

The `.finally(() => setSettledFlushes(count => count + 1))` at this line has no generation check. `resetProcessingCounters` resets both counters to 0, but a `compressImage` call already in flight will unconditionally increment `settledFlushes` to 1 after the reset.

**Result:** For the rest of the session, `processedFrames = totalItems - processingCount + 1` while `totalProcessingFrames = totalItems + 0`, so `processedFrames > totalProcessingFrames` — the progress bar is permanently over-counted by 1 after any discard that races a flush.

**Fix:** snapshot a `resetGeneration` counter at flush-start time (alongside the existing `flushGeneration`) and check it inside `.finally()` before calling `setSettledFlushes`, mirroring the `flushGeneration === generation.current` guard used for `onFrameSelected`.

packages/inspection-capture-web/src/VideoCapture/hooks/useFrameSelection/useFrameSelection.ts
```diff
@@ -73,22 +103,60 @@ export function useFrameSelection({
...
+  }, [handle, onFrameSelected, handleError]);
+
+  const discardBestFrame = useCallback(() => {
+    generation.current += 1;
     bestScore.current = null;
     bestFrame.current = null;
-  }, frameSelectionInterval);
+  }, []);
+
+  const resetProcessingCounters = useCallback(() => {
```
dlymonkai, line 134: **[Bug — CONFIRMED] Root cause: `resetProcessingCounters` has no generation guard — all in-flight `.finally()` callbacks remain active post-reset**

The `generation` mechanism on `discardBestFrame` correctly guards `onFrameSelected`, but `setSettledFlushes` in `.finally()` has no equivalent. `resetProcessingCounters` introduces a second 'generation' boundary (the reset itself) that in-flight callbacks must respect; without it the settled counter is poisoned after every discard that races a `compressImage` call.

This is the root cause of the counter corruption described in the finding above at line 125.

packages/inspection-capture-web/src/VideoCapture/hooks/useFrameSelection/useFrameSelection.ts
```diff
@@ -73,22 +103,60 @@ export function useFrameSelection({
...
+    setStartedFlushes(0);
+    setSettledFlushes(0);
+    processingQueue.clear();
+  }, [processingQueue.clear]);
+
+  useInterval(flushBestFrame, flushTrigger === undefined ? frameSelectionInterval : null);
```
dlymonkai, line 140: **[Bug — PLAUSIBLE] In FIXED_UPLOAD_RATE mode the flush interval keeps firing after the screen transitions to PROCESSING or COMPLETE**

`useInterval(flushBestFrame, flushTrigger === undefined ? frameSelectionInterval : null)` evaluates to a live interval in FIXED_UPLOAD_RATE mode and is never stopped when recording ends. While `bestFrame.current` is usually null by then (no new screenshots), a screenshot pushed right at the recording boundary could still be processed by the Laplace queue and set `bestFrame.current` before the next interval fires — uploading a frame into the PROCESSING phase after the recording has ended and shifting the progress bar unexpectedly.

packages/inspection-capture-web/src/VideoCapture/hooks/useFrameSelection/useFrameSelection.ts
```diff
@@ -73,22 +103,60 @@ export function useFrameSelection({
...
+
+  const flushBestFrameRef = useRef(flushBestFrame);
+  flushBestFrameRef.current = flushBestFrame;
+  const lastFlushTrigger = useRef(flushTrigger);
```
dlymonkai, line 144: **[Bug — PLAUSIBLE] `lastFlushTrigger` not reset on new walkaround — rapid batched bucket crossings drop a segment**

If two bucket crossings happen in rapid succession (batched React state updates), `setFlushTrigger` increments twice before the `useEffect` re-runs. The effect fires once, advances `lastFlushTrigger.current` by 2, and calls `flushBestFrame()` once — silently dropping the intermediate bucket's best frame entirely.

In React StrictMode (dev) the double-invocation is harmless, but the batched case is a real data loss on fast walkarounds.

packages/inspection-capture-web/src/VideoCapture/VideoCaptureHUD/VideoCaptureHUD.tsx
```diff
@@ -153,17 +165,53 @@ export function VideoCaptureHUD({
...
   });
+
+  const handleStartWalkaround = useCallback(() => {
+    discardBestFrame();
+    startWalkaround();
+    startSegmentTracking();
+  }, [discardBestFrame, startWalkaround, startSegmentTracking]);
+
+  const handleDiscardVideo = useCallback(() => {
+    discardBestFrame();
```
dlymonkai, line 196: **[Bug — CONFIRMED] `discardBestFrame()` called twice on discard — generation incremented an extra time**

The discard dialog is only reachable after `pauseRecording()` sets `isRecordingPaused=true`. A `useEffect` watching `isRecordingPaused` already calls `discardBestFrame()` (generation → G+1) to abort in-flight uploads. `handleDiscardVideo` then calls it again (generation → G+2).

The extra increment means any flush started right after `resetProcessingCounters` (which does not reset `generation`) but before the next recording begins will be silently suppressed as stale — even though it belongs to the new recording.

**Fix:** remove `discardBestFrame()` from `handleDiscardVideo`; the effect already handles it.

packages/inspection-capture-web/src/VideoCapture/hooks/useSegmentFrameSelection/useSegmentFrameSelection.ts
```diff
@@ -0,0 +1,114 @@
...
+export function useSegmentFrameSelection({
+  walkaroundPosition,
+  isRecording,
+  targetFramesCount,
+}: UseSegmentFrameSelectionParams): SegmentFrameSelectionHandle {
+  const totalBuckets = useMemo(
+    () => clampTargetFramesCount(targetFramesCount),
+    [targetFramesCount],
+  );
+  const bucketSizeDegrees = 360 / totalBuckets;
```
dlymonkai, line 84: **[Efficiency — PLAUSIBLE] `bucketSizeDegrees` is unmemoized but listed in `useEffect` deps**

`const bucketSizeDegrees = 360 / totalBuckets` is a plain variable derived from the memoized `totalBuckets` and listed as a `useEffect` dependency. Currently safe — `Object.is()` float comparison holds as long as `totalBuckets` is an integer. But if the computation ever produces a new floating-point value per render (e.g. if `totalBuckets` becomes a float), the effect — and thus `setFlushTrigger` — would fire on every position update even when the bucket has not changed.

**Fix:** inline `360 / totalBuckets` directly in the effect body (eliminating the dependency), or wrap `bucketSizeDegrees` in `useMemo`.

packages/inspection-capture-web/src/VideoCapture/hooks/useSegmentFrameSelection/useSegmentFrameSelection.ts
```diff
@@ -0,0 +1,114 @@
...
+    const currentBucket = Math.floor(walkaroundPosition / bucketSizeDegrees) % totalBuckets;
+    if (capturedBuckets.current.has(currentBucket)) {
+      return;
+    }
+    capturedBuckets.current.add(currentBucket);
+    setCapturedFramesCount(capturedBuckets.current.size);
+    setFlushTrigger((value) => value + 1);
+  }, [walkaroundPosition, isRecording, bucketSizeDegrees, totalBuckets]);
+
+  const startSegmentTracking = useCallback(() => {
```
dlymonkai, line 103: **[Bug — CONFIRMED] `startSegmentTracking` pre-seeds capturedFramesCount=1 without a flush — stop button lights up after 39 uploads instead of 40**

`startSegmentTracking` sets `capturedBuckets={0}` and implicitly makes `capturedFramesCount=1`, without triggering `setFlushTrigger`. Buckets 1–39 then fire 39 segment flushes, bringing `capturedFramesCount` to 40 → `isCaptureComplete` returns `true` → stop button activates. But only **39 frames** have been uploaded via segment tracking; the 40th upload happens inside `handleRecordingComplete` only after the user taps Stop.

Additionally, `capturedFramesCount=1` is surfaced to the UI immediately at recording start, falsely showing 1 frame captured before any upload occurs.

**Fix:** don't pre-seed bucket 0 in `startSegmentTracking`; let the effect capture it naturally on the first position update. This keeps the count accurate and re-aligns the 'complete' signal with actual uploads.

packages/inspection-capture-web/src/VideoCapture/hooks/useVideoRecording.ts
```diff
@@ -255,6 +335,7 @@ export function useVideoRecording({
     onDiscardDialogKeepRecording,
     onDiscardDialogDiscardVideo,
     isDiscardDialogDisplayed,
+    isMissingTargetFrames,
```
dlymonkai, line 338: **[Bug — PLAUSIBLE] `isMissingTargetFrames` not reset on discard — stale state between sessions**

`onDiscardDialogDiscardVideo` (line 279) does not reset `isMissingTargetFrames` to `false`. If the user triggered the 'missing frames' dialog, then completed a successful recording (no dialog shown), the flag stays `true`. In a subsequent session, `onClickRecordVideo` overwrites the flag on the same click event before showing the dialog — but there is a brief render window where the wrong translation key is passed to `t()`, potentially causing a flash of the wrong message.

**Fix:** add `setIsMissingTargetFrames(false)` inside `onDiscardDialogDiscardVideo`.

dlymonkai (commented):
Third review pass — findings on the latest commit.

dlymonkai (approved):

